PR #13, opened 2025-09-25
VCV-5: Add 3 dropdowns to the annotation dashboard - Species, Sex, Abdomen Status - with the ability to clear selection

vinnychi123:
I created a select compononent in the annotation-form-panel form that took in a string arrray of species, sexes, and abdomen statuses, and created select menus for each. 

When I originally tested it, the options would remain there after I went to a new page, but I added useEffect to set them all the undefined (nothing selected) every page change--this may not be the most optimal at the moment, but I think it's still better than leaving it unchanged.


aryamanshodhan (changes requested):
Good job Vincent! Please address the comments in this PR. There are several opportunities for code consolidation and optimization. Your approach makes a lot of sense though. From a testing perspective, I see a bit of flicker for the species dropdown, not sure why that is happening. Also, when I hover over "Remove Selection", the background turns red, but the text turns black. Can we keep the text red when this happens?

Dont hesitate to be as modular as possible, and create files as needed! :)

src/lib/entities/specimen/morph-ids.ts
```diff
@@ -0,0 +1,21 @@
+export const MorphIds = {
```
aryamanshodhan, line 1: SPECIES_MORPH_IDS

src/lib/entities/specimen/morph-ids.ts
```diff
@@ -0,0 +1,21 @@
...
+    ANOPHELES_FUNESTUS: "Anopheles funestus",
+    ANOPHELES_BAMBIAE: "Anopheles gambiae",
+    ANOPHELES_OTHER: "Anopheles other",
+    CULEX: "Culex",
+    AEDES: "Aedes",
+    MANSIONIA: "Mansonia",
+    NON_MOSQUITO: "Non-Mosquito"
+}
+
+export const SexMorphIds = {
```
aryamanshodhan, line 11: SEX_MORPH_IDS

src/lib/entities/specimen/morph-ids.ts
```diff
@@ -0,0 +1,21 @@
...
+    MALE: "Male",
+    FEMALE: "Female"
+}
+
+export const AbdomenStatusMorphIds = {
```
aryamanshodhan, line 16: ABDOMEN_STATUS_MORPH_IDS

aryamanshodhan (changes requested):
This is ready to ship. Need a last variable name change. Constants should be defined with all caps variable names!

aryamanshodhan (approved):
Ready to ship! Well done Vincent

aryamanshodhan (approved):
Looks good to me!

PR #15, opened 2025-09-26
VCV-1: Add a filter by relative date to the Annotations List Page

jyoon53:
- Local state management - Date range selection (1-month, 3-months, etc.) is stored in component state, with automatic conversion to ISO date strings for backend queries Backend filtering 
- The actual data filtering is handled by the BFF using fromDate/toDate query parameters, so the server does the heavy lifting of filtering annotation tasks by date range Automatic pagination reset 
- When users change the date filter, pagination automatically resets to page 1 since the filtered dataset changes, ensuring users see results from the beginning of the new date range

### DIFF
--- a/src/components/annotate/tasks-list/date-range-filter.tsx
+++ b/src/components/annotate/tasks-list/date-range-filter.tsx
@@ -1,0 +1,48 @@
+'use client';
+
+import {
+  Select,
+  SelectContent,
+  SelectItem,
+  SelectTrigger,
+  SelectValue,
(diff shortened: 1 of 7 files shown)


aryamanshodhan (approved):
Looks great Kurtis! This is ready to merge. HOWEVER, lets hold off on the merge until the backend changes are made. Then lets test it out!

vinnychi123 (approved):

aryamanshodhan (approved):
Great job Kurtis! Looks great

PR #17, opened 2025-09-28
VCV-8: Add a notes field, flag specimen button, and submit button

vinnychi123:
Added notes section, flag and save buttons. No validation, but the buttons are there and I can type in notes

I think I need build the validation schema for the annotation form before I can move from card to form, but I plan to do that with VCV-7


aryamanshodhan (changes requested):
This PR needs to be closed. Please reopen another one where you merge into develop, not main!

PR #18, opened 2025-09-28
VCV-8: Add a notes field, flag specimen button, and submit button

vinnychi123:
Added notes section, flag and save buttons. No validation, but the buttons are there and I can type in notes

I think I need build the validation schema for the annotation form before I can move from card to form, but I plan to do that with VCV-7

### DIFF
--- a/src/components/annotate/task-detail/page-client.tsx
+++ b/src/components/annotate/task-detail/page-client.tsx
@@ -14,6 +14,8 @@ import {
   useAnnotationTaskProgressQuery,
   useTaskAnnotationsQuery,
 } from '@/lib/annotate/client';
+import { Flag, Save } from 'lucide-react';
+import { Textarea } from '@/components/ui/textarea';
 import { formatDate } from '@/lib/shared/utils/date';
 import { ArrowLeft, ArrowRight, CalendarDays, Info } from 'lucide-react';
 import { useCallback, useEffect, useMemo, useState } from 'react';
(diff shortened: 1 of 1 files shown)


src/components/annotate/task-detail/page-client.tsx
```diff
@@ -194,7 +196,36 @@ export function AnnotationTaskDetailPageClient({
...
+            </div>
+
+            <div className = "flex gap-3 pt-2">
+              <Button 
+                type="button"
+                variant="outline"
+                className="flex-1"
+                disabled={isFetching || isLoading}
+              >
+                <Flag className="mr-2 h-4 w-4" /> Flag Species
```
aryamanshodhan, line 214: [Suggestion]: You should replace this with "Flag Specimen", because you are flagging the whole specimen, not just the species.

src/components/annotate/task-detail/page-client.tsx
```diff
@@ -194,7 +196,36 @@ export function AnnotationTaskDetailPageClient({
...
+            <div className="space-y-2">
+              <h3 className="text-foreground text-sm font-bold">Notes</h3>
+              <Textarea
+                className="h-[70px] resize-none text-sm overflow-y-auto"
+                placeholder="Add observations..."
+              />
+            </div>
+
+            <div className = "flex gap-3 pt-2">
+              <Button
```
aryamanshodhan, line 208: [Suggestion]: I think this should be a toggle component, because you are toggling the flagged and not flagged state. Check out the UI on the login branch. Also, currently when I click the button, I get no feedback that I have flagged the specimen. The older UI made it flash and turn red. Check out the code on the login branch. Run the code on the login branch and check out the UI yourself and play around with it.

src/components/annotate/task-detail/page-client.tsx
aryamanshodhan, line 1: Question: I am guessing you will handle the form state and the value updates in VCV-7? On the login branch, properly check out the AnnotationForm component, and see how the validation schema and "watching" of value changes happens. Also down to do a paired programming session if that is more efficient.

aryamanshodhan (changes requested):
Nice work Vincent! Check out my comments, and please submit an updated PR.

aryamanshodhan (approved):
Well done Vincent! This looks good in my opinion. 

I would suggest an optional change, but I think it will be addressed in VCV-7: make sure to set isFlagged to false whenever the user clicks next or previous.

PR #19, opened 2025-09-28
VCV-7: Add Zod validation to the dropdowns, notes, and flag specimen button

vinnychi123:
created validation so that if the specimen is not flagged -- species, sex, and abdomen status select menu must be selected.

if specimen is flagged, there must be something in notes box.

switched out cards for forms in order to implement zod validation, which can be found in components/annotate/task-detail/annotation-form-panel/validation. if user clicks clear select, validation disappears. and if select menu is blanked out, there will be no validation.


jyoon53 (approved):
Looks good!

PR #22, opened 2025-10-03
VCV-17: Permission-based rendering of a district filter for data review list items

jyoon53:
…ew list items


aryamanshodhan (changes requested):
Good job Kurtis! This seems to work as expected. I have left some comments for you to address. I wonder if you can apply a similar approach to the DateRange filter for this PR? This code seems to be a lot more complex than how you had implemented the Date Range filter.

PR #23, opened 2025-10-03
VCV-28: Dashboard Permission Fix: Enables Review Access for Privilege 1 Users

jyoon53:
- Fixed blank dashboard issue - Separated annotation and review card permission checks so privilege 1 users can access review functionality
- Maintained security boundaries - Annotation card remains restricted to privilege 2 users while review card is accessible to all users with site viewing permissions
- Preserved code architecture - Solution follows existing patterns with proper separation of concerns, single responsibility principle, and clean file structure

✅ Privilege 0: Can access review (read-only for single district) - canViewSites() returns true for permissions.sites.viewSiteMetadata
✅ Privilege 1: Can access review (read/write/push for single district) - canViewSites() returns true for permissions.sites.viewSiteMetadata
✅ Privilege 2: Can access review (full access across districts) - canViewSites() returns true for permissions.sites.viewSiteMetadata

### DIFF
--- a/src/components/dashboard/dashboard-page-client.tsx
+++ b/src/components/dashboard/dashboard-page-client.tsx
@@ -1,7 +1,10 @@
 'use client';
 
 import { useUserPermissionsQuery } from '@/lib/user/client';
-import { canViewAndWriteAnnotationTasks } from '@/lib/user/permissions';
+import {
+  canViewAndWriteAnnotationTasks,
+  canViewSites,
+} from '@/lib/user/permissions';
 import { Button } from '../ui/button';
 import {
   Card,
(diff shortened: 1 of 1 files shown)


src/components/dashboard/dashboard-page-client.tsx
```diff
@@ -27,87 +30,93 @@ export function DashboardPageClient() {
     router.push('/review');
   };
 
+  const canAnnotate =
+    permissions && canViewAndWriteAnnotationTasks(permissions);
+  const canReview = permissions && canViewSites(permissions);
+  const hasAnyAccess = canAnnotate || canReview;
+
+  if (!hasAnyAccess) {
```
aryamanshodhan, line 38: If they have no access, instead of null, it might be better to return something to avoid them needing to stare at a white screen :)

aryamanshodhan (changes requested):
Looks good Kurtis, one minor change needed!

PR #25, opened 2025-10-05
VCV-27: Rethink pagination

jyoon53:
- made it more concise. 
- replaced previous and next with first and last. 
- Got rid of left ellipse. 
- only shows the current page and neighboring page options.


aryamanshodhan (approved):
Looks good Kurtis!

PR #43, opened 2025-11-06
VCV-44: Submitting an annotation should send the user to the closest next unannotated image.

vinnychi123:
Needed to include another API call in order to implement logic so that previous forms could be traversed to previous pages that haven't still have a "pending" status.

### DIFF
--- a/src/features/annotation/components/task-detail/page-client.tsx
+++ b/src/features/annotation/components/task-detail/page-client.tsx
@@ -32,6 +32,7 @@ export function AnnotationTaskDetailPageClient({
   taskId,
 }: AnnotationTaskDetailPageClientProps) {
   const [page, setPage] = useState(1);
+  const [statusFilter, setStatusFilter] = useState<'PENDING' | undefined>(undefined);
   const queryClient = useQueryClient();
 
   const {
@@ -42,6 +43,7 @@ export function AnnotationTaskDetailPageClient({
     taskId,
     page,
     limit: 1,
+    status: statusFilter,
   });
 
   const { data: taskProgress } = useAnnotationTaskProgressQuery(taskId);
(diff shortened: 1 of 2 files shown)


src/features/annotation/components/task-detail/page-client.tsx
```diff
@@ -75,6 +79,83 @@ export function AnnotationTaskDetailPageClient({
...
+      });
+
+      if (pendingAnnotations.items && pendingAnnotations.items.length > 0) {
+        firstPendingId = pendingAnnotations.items[0].id;
+        
+        let found = false;
+        let currentBatch = 1;
+        const batchSize = 50;
+        
+        while (!found && currentBatch <= 10) {
```
aryamanshodhan: @Vincent Can you please explain the purpose of this while loop?

src/features/annotation/components/task-detail/page-client.tsx
```diff
@@ -75,6 +79,83 @@ export function AnnotationTaskDetailPageClient({
...
+            found = true;
+          } else if (!batch.hasMore) {
+            if (hasMore) setPage(prev => prev + 1);
+            break;
+          }
+          
+          currentBatch++;
+        }
+        
+        if (!found && hasMore) {
```
aryamanshodhan: Why do you need to run through each page? I wonder if there is a simpler way for you to query which is the next pending ID, and then just navigate to that one. Think about whether you can use the response from the backend to derive which page you should navigate to as opposed to running through a while loop.

aryamanshodhan (changes requested):

aryamanshodhan:
Additionally, I noticed something:
Suppose I have specimens 1, 2, 3, 4, 5, 6
If I annotate 1, 2, skip 3, then annotate and submit 4, it takes me back to 3. Then when I go to 5, it takes me back to 3. I think it is more user friendly to prioritize the next image. Just because there was probably a reason why the user decided to skip the image previously, so maybe might want to do it in the end. But if after every specimen I keep getting redirected to that specimen, it might be frustrating for them.

jyoon53 (dismissed):
Looks good to me. It works.

src/features/annotation/components/task-detail/page-client.tsx
```diff
@@ -75,6 +79,48 @@ export function AnnotationTaskDetailPageClient({
...
+      if (hasMore) {
+        setPage(prev => prev + 1);
+      }
+    } catch (error) {
+      console.error('Error in handleAnnotationSuccess:', error);
+      if (hasMore) {
+        setPage(prev => prev + 1);
+      }
+    }
+  }, [page, taskId, hasMore, isFetching, isLoading]);
```
jyoon53: Why is isFetching necessary?

src/features/annotation/components/task-detail/page-client.tsx
```diff
@@ -75,6 +79,48 @@ export function AnnotationTaskDetailPageClient({
     }
   }, [page, isFetching, isLoading]);
 
+  const handleAnnotationSuccess = useCallback(async () => {
+    if (isFetching || isLoading) return;
```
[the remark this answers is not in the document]
aryamanshodhan: isLoading is true only during the very first fetch, before any data exists in the cache—typically on initial mount or when the query key changes and there’s no cached result to reuse.

isFetching is true whenever the query function is running, even if you already have cached data. That includes background refetches triggered by window focus, retry cycles, or manual refetch(). It’s false only when the network is idle.

aryamanshodhan (changes requested):
Well done Vincent. i think there are some opportunities for clean up, as well as potential bugs. Please review comments from everyone and resolve them. Thanks!

PR #50, opened 2025-11-18
VCV-67: Annotation Form Page UI

vinnychi123:
Added a genus-morph-id and options to choose a specimen genus. Also recolored ui to match vectorcam colors.


aryamanshodhan (changes requested):
This is good Vincent. A couple of points:
1. What if there are multiple artifacts (Eg. 2 specimens + blurry)? Can you make it a multiple select? I think shadcn allows that. 
2. Abdomen status is disabled by default before I select a sex. Is this something we want to do? Or just disable it after the user selects non mosquito or male for the sex.
3. VCV-49 handles the morph ID and visual ID panels, so I think this PR will require significant changes after that gets merged in. Because currently we do not hanfle morph ID and visual ID separately. Ultimately, this PR is blocked before VCV-49 is merged in.

jyoon53:
The option to choose specimen genus works for visual ID form and its drop downs, but it doesn't work for morph ID form. We can also make a separate task for this. But besides this, I think everything works fine.

src/features/annotation/components/task-detail/annotation-form-panel/visual-identification-form.tsx
```diff
@@ -49,13 +48,24 @@ interface AnnotationFormProps {
...
 
-export function AnnotationForm({
+export function VisualIdentificationForm({
   annotationId,
   defaultValues,
   onGenusChange,
-}: AnnotationFormProps) {
+  morphFormValues,
+  shouldProcessFurther = false,
+  morphFormRef,
```
aryamanshodhan, line 63: Question: Why is morph form data inside visual identification form? I also do not see them being used anywhere

aryamanshodhan (changes requested):
Nice job Vincent. 

i noticed one issue. the form does not preload values if the image is already annotated. I have attached an example. In the image below, you can see that the image was annotated, but I cannot see what the selections were.

Image

src/features/annotation/components/task-detail/annotation-form-panel/visual-identification-form.tsx
vinnychi123: since the submit is in visual identification form, I need the info from the morph id form in order to have a valid submit.

aryamanshodhan (changes requested):
Vincent I have a question. How are you handling this situation? If an existing annotated image is changed to flagged without changing any of the selections for the visual IDs, when I reload the page, I am not able to see the selected species. When the specimen is flagged, are you clearing the species, sex, abdomen status selections? 

Image

Image

aryamanshodhan (approved):
Looks good Vincent, well done. Please add this to the excel sheet before merging in. Thanks!

PR #57, opened 2025-11-22
VCV-78: Add query parameters to show data for only surveillance sessions and specimens

vinnychi123:
added type; "SURVEILLANCE" as query when fetching sessions.


aryamanshodhan (dismissed):

jyoon53 (approved):

aryamanshodhan (changes requested):
The review list page still shows sessions from data collection mode. Not sure if this is the scope of this task. Please add a description of changes to the excel sheet.

aryamanshodhan (approved):

PR #66, opened 2026-02-02
VCV-96: annotation panel shows morph ID annotation form for every specimen

jyoon53:
VCV-96: Morph form is no longer gated by specimen.shouldProcessFurther. Removed useShouldProcessFurther hook, getSpecimen API, and related conditionals.


src/features/annotation/components/task-detail/constants.ts
aryamanshodhan, line 1: Not sure if this is an appropriate file to create CSS constants. I think you can do it within the file you are inserting this CSS into.

next.config.ts
```diff
@@ -1,7 +1,7 @@
 import type { NextConfig } from 'next';
 
 const nextConfig: NextConfig = {
-  /* config options here */
+  transpilePackages: ['@radix-ui/react-hover-card'],
```
aryamanshodhan: Why do we need this?

aryamanshodhan (changes requested):
Hey Kurtis! This is good. Just address the comments left here, and please do another round of testing since the annotation form UI was updated recently.

vinnychi123 (dismissed):

aryamanshodhan (approved):

PR #67, opened 2026-02-04
VCV-104: Default review entry to Master Table View

jyoon53:
- Review data list: row action goes to Master Table View (was metrics dashboard)
- Add buildMasterTableViewPath; list uses it for navigation
- Sidebar: Master Table View first, then Metrics Dashboard
- List button: label "View Table", Table icon (was "Dashboard", BarChart3)
- Sidebar: label "Metrics Dashboard" (was "Dashboard")

### DIFF
--- a/src/features/review/components/data-list/page-client.tsx
+++ b/src/features/review/components/data-list/page-client.tsx
@@ -12,7 +12,7 @@ import { createMonthlySummaryColumns } from './columns';
 import { useMonthlySummaryQuery } from '@/features/review/hooks/use-monthly-summary';
 import { useTablePagination } from '@/shared/core/hooks/use-table-pagination';
 import { useDistrictManagement } from '@/features/review/hooks/use-district-management';
-import { buildDashboardPath } from '@/features/review/utils/navigation';
+import { buildMasterTableViewPath } from '@/features/review/utils/navigation';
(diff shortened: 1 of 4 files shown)


src/features/review/components/data-list/page-client.tsx
```diff
@@ -82,7 +82,7 @@ export function ReviewDataListPageClient() {
 
   const handleNavigateToDashboard = useCallback(
     (district: string, monthYear: string) => {
-      router.push(buildDashboardPath(district, monthYear));
+      router.push(buildMasterTableViewPath(district, monthYear));
```
aryamanshodhan, line 85: If buildDashboardPath is no longer being used, should we remove it?

aryamanshodhan (commented):
Looks good. Only one question as part of this PR.

src/features/review/components/data-list/page-client.tsx
vinnychi123: We should also rename to handleNavigateToMasterTableView

vinnychi123 (approved):

aryamanshodhan (approved):
This looks good Kurtis! I have a question: Should we move the sync to DHIS2 button into this screen instead of the dashboard?

PR #69, opened 2026-02-12
VCV-102: Session view items should show location where data was collected from

vinnychi123:
- Added endpoint that returns address of each session.
- Displayed address in the accordion content of every session.
- Also refactored session data


aryamanshodhan (dismissed):
Looks good Vincent!

PR #73, opened 2026-03-10
VCV-130: Operations page should display list of cards for each district the user has access to and wire into the /operations/[district] page

jyoon53:
created card component for each district.


src/api/session/validation/get-sessions-schema.ts
```diff
@@ -10,7 +10,7 @@ export const getSessionsQueryParamsSchema = z.object({
     collectorName: z.string().optional(),
     collectionMethod: z.string().optional(),
     specimenCondition: z.string().optional(),
-    status: z.enum(['pending', 'completed', 'submitted']).optional(),
+    state: z.enum(['NEEDS_REVIEW', 'IN_REVIEW', 'CERTIFIED', 'SUBMITTED', 'NOT_APPLICABLE']).optional(),
```
csmanel: Since we are defining the same enum in sessionState, could we do something like this?

```
export const sessionStateSchema = z.enum(['NEEDS_REVIEW', 'IN_REVIEW', 'CERTIFIED', 'SUBMITTED', 'NOT_APPLICABLE']);
```

and then reuse just `state: sessionStateSchema...`

vci-developers: I second this, but i think this already exists in VCV-120 i believe

src/app/(dashboard)/operations/[district]/page.tsx
```diff
@@ -0,0 +1,12 @@
+interface OperationsDistrictPageProps {
+    params: Promise<{ district: string }>;
+}
+
+export default async function OperationsDistrictPage({
+    params,
+}: OperationsDistrictPageProps) {
+    const { district } = await params;
+    const decodedDistrict = decodeURIComponent(district);
```
Comment (vci-developers) [inline src/app/(dashboard)/operations/[district]/page.tsx]: Question: Should we merge these 2 lines maybe? decodedDistrict is a kinda cryptic variable name

const district = decodeUriComponent((await params).district)

src/features/operations/components/district-list/operations-district-card.tsx
vci-developers, line 1: In a subsequent PR, we need to replace usage of district with the highest site level within locationType. Please add this as a task to JIRA

src/features/operations/components/district-list/operations-district-list-page-client.tsx
vci-developers, line 1: Same here. We need to be generic. Let's avoid the use of district in the file name

src/features/operations/components/district-list/operations-district-list-page-client.tsx
```diff
@@ -0,0 +1,28 @@
+'use client';
+
+import { useDistrictList } from '@/features/operations/hooks/use-district-list';
+import PageShell from '@/components/layout/page-shell';
+import { Activity } from 'lucide-react';
+import OperationsDistrictList from './operations-district-list';
+
+export default function OperationsDistrictListPageClient() {
+    const { districts, isPending, isError } = useDistrictList();
```
vci-developers: I would suggest implementing this logic inline as opposed to making it a hook. The hook is way too specific, especially since it is being used in just one file.

src/features/operations/components/district-list/operations-district-list-page-client.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+    return (
+        <PageShell
+            title="Operations"
+            description="District-level entomological metrics and session data"
+            icon={Activity}
+        >
+            {isPending ? (
```
vci-developers: In order to be consistent with other files, lets early return from the function for loading and error states

src/features/operations/hooks/use-district-list.ts
```diff
@@ -0,0 +1,43 @@
...
+    siteCount: number;
+}
+
+interface UseDistrictListResult {
+    districts: DistrictInfo[];
+    isPending: boolean;
+    isError: boolean;
+}
+
+export function useDistrictList(): UseDistrictListResult {
```
vci-developers: I would recommend having the logic in this file be implemented within the component itself, because this implementation is only used in one component, so I dont think we need to abstract it at this point.

src/features/operations/hooks/use-district-list.ts
```diff
@@ -0,0 +1,43 @@
...
+    const { data: permissionsResult, isPending } = useGetUserPermissions();
```
vci-developers: getUserPermissionsResult, isGetUserPermissionsPending

src/features/operations/components/district-list/operations-district-list.tsx
```diff
@@ -0,0 +1,32 @@
...
+import OperationsDistrictCard from './operations-district-card';
+
+interface OperationsDistrictListProps {
+    districts: DistrictInfo[];
+}
+
+export default function OperationsDistrictList({
+    districts,
+}: OperationsDistrictListProps) {
+    if (districts.length === 0) {
```
vci-developers: Maybe this can be moved into the parent component. Maybe we can display the empty state with the loading, and error states. The list can be displayed for the success state.

src/features/operations/components/district-list/operations-district-list.tsx
vci-developers, line 1: This file should be renamed to something more generic. Remove the use of district from the name

vci-developers (commented):
Great job kurtis! A few minor comments for this PR, otherwise it looks good!

csmanel:
Operations page has been addressed in current VCV-97

PR #74, opened 2026-03-11
VCV-133: Add to the signup form fields for entering the user's name and selecting a program

csmanel:
[VCV-133](https://vectorcam.atlassian.net/browse/VCV-133?atlOrigin=eyJpIjoiNTM0NmQzMjJmZDFjNDY0OWE3NTFmNmQ0OWFkZTM4ZmYiLCJwIjoiaiJ9) : Added `name` and `programId` fields to signup form 
- `signup-form.tsx` : added Name and Program fields
- `signup-form-schema.ts` : added `name` + `programId` validation
- `signup-network-schema.ts` : added `name` + `programId` to request schema
- `signup/page.tsx` : created server component to fetch programs and pass as prop 
- `get-programs.ts / get-programs-schema.ts` : new files to fetch and validate `/programs` endpoint in order to populate programs dropdown

### DIFF
--- a/src/api/program/get-programs.ts
+++ b/src/api/program/get-programs.ts
@@ -4,7 +4,7 @@ import type { Result } from '@/lib/result/result';
 import {
     getProgramsResponseSchema,
     type GetProgramsResponseBody,
-} from './validation/get-programs-schema';
+} from '@/api/program/validation/get-programs-schema';
 
 export async function getPrograms(): Promise<
     Result<GetProgramsResponseBody, NetworkError>
--- a/src/api/program/get-programs.ts
+++ b/src/api/program/get-programs.ts
@@ -2,15 +2,23 @@ import type { NetworkError } from '@/lib/network/network-error';
 import { safeApiCall } from '@/lib/network/safe-api-call';
 import type { Result } from '@/lib/result/result';
 import {
+    getProgramsQueryParamsSchema,
     getProgramsResponseSchema,
+    type GetProgramsQueryParams,
     type GetProgramsResponseBody,
 } from '@/api/program/validation/get-programs-schema';
+import { constructQueryString } from '@/lib/network/construct-query-string';
 
-export async function getPrograms(): Promise<
-    Result<GetProgramsResponseBody, NetworkError>
-> {
+export async function getPrograms(
+    queryParams?: GetProgramsQueryParams,
+): Promise<Result<GetProgramsResponseBody, NetworkError>> {
+    const queryString = constructQueryString<GetProgramsQueryParams>(
+        queryParams,
+        getProgramsQueryParamsSchema,
+    );
+
     return safeApiCall<GetProgramsResponseBody>(
-        '/programs',
+        `/programs${queryString}`,
         {
             method: 'GET',
         },
--- a/src/api/program/get-programs.ts
+++ b/src/api/program/get-programs.ts
@@ -1,0 +1,19 @@
+import type { NetworkError } from '@/lib/network/network-error';
+import { safeApiCall } from '@/lib/network/safe-api-call';
+import type { Result } from '@/lib/result/result';
+import {
+    getProgramsResponseSchema,
+    type GetProgramsResponseBody,
+} from './validation/get-programs-schema';
+
+export async function getPrograms(): Promise<
+    Result<GetProgramsResponseBody, NetworkError>
+> {
+    return safeApiCall<GetProgramsResponseBody>(
+        '/programs',
+        {
+            method: 'GET',
+        },
+        getProgramsResponseSchema,
+    );
+}
--- a/src/api/program/validation/get-programs-schema.ts
+++ b/src/api/program/validation/get-programs-schema.ts
@@ -9,4 +9,12 @@ export const getProgramsResponseSchema = z.object({
     hasMore: z.boolean(),
 });
 
+export const getProgramsQueryParamsSchema = z.object({
+    programId: z.coerce.number().optional(),
+    name: z.string().optional(),
+    country: z.string().optional(),
+    limit: z.coerce.number().min(1).max(100).optional(),
+    offset: z.coerce.number().min(0).optional(),
+});
+
 export type GetProgramsResponseBody = z.infer<typeof getProgramsResponseSchema>;
--- a/src/api/program/validation/get-programs-schema.ts
+++ b/src/api/program/validation/get-programs-schema.ts
@@ -18,3 +18,6 @@ export const getProgramsQueryParamsSchema = z.object({
 });
 
 export type GetProgramsResponseBody = z.infer<typeof getProgramsResponseSchema>;
+export type GetProgramsQueryParams = z.infer<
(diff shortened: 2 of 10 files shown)


src/api/program/validation/get-programs-schema.ts
aryamanshodhan, line 1: Question: I know we don't use it at this point, but should we add the queryParams filters for GET programs?

src/api/program/get-programs.ts
```diff
@@ -0,0 +1,19 @@
+import type { NetworkError } from '@/lib/network/network-error';
+import { safeApiCall } from '@/lib/network/safe-api-call';
+import type { Result } from '@/lib/result/result';
+import {
+    getProgramsResponseSchema,
+    type GetProgramsResponseBody,
+} from './validation/get-programs-schema';
```
aryamanshodhan: Let's make all the imports relative to the src folder (@/api/programs/validation/get-programs-schema) instead of relative imports from the current directory.

src/app/(auth)/signup/page.tsx
```diff
@@ -1,16 +1,20 @@
+import { getPrograms } from '@/api/program/get-programs';
 import { Separator } from '@/components/ui/separator';
 import AuthShell from '@/features/auth/components/auth-shell';
 import SignupForm from '@/features/auth/components/signup-form';
 import Link from 'next/link';
 
-export default function SignupPage() {
+export default async function SignupPage() {
+    const programsResult = await getPrograms();
```
aryamanshodhan: Question: Should we fetch the list of programs in the SignupForm component?

aryamanshodhan: getProgramsResult

src/app/(auth)/signup/page.tsx
```diff
@@ -1,16 +1,20 @@
...
+    const programs = programsResult.ok ? programsResult.data.programs : [];
```
aryamanshodhan: If the list of programs could not be fetched, maybe we should display an error banner instead of having an empty list? Maybe have the error condition: 

if (!getProgramsResult.ok) {
    // Return error JSX
}

src/features/auth/validation/form/signup-form-schema.ts
```diff
@@ -3,9 +3,15 @@ import { z } from 'zod';
 export const signupFormSchema = z
     .object({
         email: z.email('Enter a valid email'),
+        name: z
+            .string()
+            .min(1, 'Name is required')
+            .max(128, 'Name is too long'),
+        programId: z.string().min(1, 'Program is required'),
```
aryamanshodhan: ProgramId should be a number. Maybe you could coerce to a number instead of having it be a string.

src/features/auth/components/signup-form.tsx
```diff
@@ -30,6 +42,8 @@ export default function SignupForm() {
         resolver: zodResolver(signupFormSchema),
         defaultValues: {
             email: '',
+            name: '',
+            programId: '',
```
aryamanshodhan: Question: Is it possible for this to be a number?

src/features/auth/components/signup-form.tsx
```diff
@@ -42,14 +56,17 @@ export default function SignupForm() {
                 'Content-Type': 'application/json',
             },
             credentials: 'include',
-            body: JSON.stringify(data),
+            body: JSON.stringify({
+                ...data,
+                programId: parseInt(data.programId),
```
aryamanshodhan: I think in your schema you can use the "coerce" keyword to check numbers, and then you might not need to use parseInt.

aryamanshodhan (changes requested):
Great job Cole! This is super clean! Just a couple of very minor comments.

src/api/program/validation/get-programs-schema.ts
csmanel: Missed that! I think it makes sense to include that now.

src/app/(auth)/signup/page.tsx
csmanel: I think as it stands, moving the fetch to the SignupForm would create unnecessary complexity? If the form needed to be reused elsewhere, it would make sense to make it self contained, but for the time being I am not sure it needs to be?

src/app/(auth)/signup/page.tsx
csmanel: Let me know if what I pushed is an okay temporary fix 

At the moment if it errors it will look like this

Image

Potentially obviously I would like it to just error on the programs selector, but I think I could create a new issue in jira for changing that UI?

src/features/auth/components/signup-form.tsx
csmanel: This relates to the following comments as well ([R10](https://github.com/vci-developers/vector-verify/pull/74/changes/BASE..22bcffad1e61f6fd7b0e38db9ab9581075be3a97#r2928300821), [R61](https://github.com/vci-developers/vector-verify/pull/74/changes/BASE..22bcffad1e61f6fd7b0e38db9ab9581075be3a97#r2928290321))

I tried to use coerce but since we are using select for the program selection it seems like no matter what we have some strange workaround? 
Here is what I tried 
- Changing `SignupFormInput` to use `z.input<>`fixed the `useForm` generic but broke the `onSubmit `handler type
- Using `useForm<SignupFormInput, unknown, SignupFormOutput>` resolved the submit handler but left `field.value` typed as `unknown`, breaking Select's value prop
- Casting `field.value as string` technically worked but felt like adding a complex work around that is potentially unsafe?

src/features/auth/components/signup-form.tsx
```diff
@@ -42,14 +63,14 @@ export default function SignupForm() {
                 'Content-Type': 'application/json',
             },
             credentials: 'include',
-            body: JSON.stringify(data),
+            body: JSON.stringify({ ...data }),
```
aryamanshodhan: We dont need to use the spread operator anymore, can just say JSON.stringify(data)

src/features/auth/components/signup-form.tsx
```diff
@@ -15,21 +15,42 @@ import {
...
+    SelectValue,
+} from '@/components/ui/select';
+import { useGetPrograms } from '@/api/program/hooks/use-get-programs';
 
 export default function SignupForm() {
     const router = useRouter();
     const [showPassword, setShowPassword] = useState(false);
     const [showConfirmPassword, setShowConfirmPassword] = useState(false);
 
+    const { data: getProgramsResult, isLoading: isProgramsLoading } =
```
aryamanshodhan: isGetProgramsPending

Also, use tanstack's isPending property, not isLoading. They mean 2 different things, and isPending is better to use than isLoading, because isLoading is only true for the initial fetch, not for subsequent refetches.

aryamanshodhan (dismissed):
Looks good Cole. A couple of optional nitpicks which would make things cleaner in my opinion.

aryamanshodhan (approved):

PR #80, opened 2026-03-27
VCV-123: Display all sessions in table form at step 1

vinnychi123:
- Created getAllSurveillanceForms function to retrieve all surveillance forms from a siteId.
- build a table of household information using queries passed through from review list.
- did not create row data component because sessions are displayed by column instead of row


src/api/surveillance-form/hooks/use-get-all-surveillance-forms.ts
```diff
@@ -0,0 +1,48 @@
...
+async function fetchAllSurveillanceForms(
+    queryParams: GetAllSurveillanceFormsQueryParams,
+): Promise<GetAllSurveillanceFormsQueryResult> {
+    const response = await fetch('/api/surveillance-forms/all', {
+        method: 'POST',
+        credentials: 'include',
+        headers: {
+            'Content-Type': 'application/json',
+        },
+        body: JSON.stringify(queryParams),
```
aryamanshodhan: If you are fetching data using a GET request, the data needs to be part of the query string. Typically for multiple sessionIds (Eg. if I want to fetch the forms for sessionIds 3, 4, 5), the query params is written as:
?sessionId=3&sessionId=4&sessionId=5 

constructQueryString is already able to convert array query parameters into this structure. So, if you pass in the query params as follows:
{
    sessionIds: [3, 4, 5]
}

constructQueryString will be able to convert it into the appropriate structure.

src/app/api/surveillance-forms/all/route.ts
```diff
@@ -0,0 +1,36 @@
+import { getAllSurveillanceForms } from '@/api/surveillance-form/get-all-surveillance-forms';
+import {
+    getAllSurveillanceFormsQueryParamsSchema,
+    type GetAllSurveillanceFormsResponseBody,
+} from '@/api/surveillance-form/validation/get-all-surveillance-forms-schema';
+import { err } from '@/lib/result/result';
+import { withAuthSession } from '@/lib/auth-session/with-auth-session';
+import { NextResponse } from 'next/server';
+
+export async function POST(request: Request) {
```
aryamanshodhan: This cannot be a POST endpoint. You are still fetching data from the backend. Therefore, it needs to be a GET endpoint. You can get the list of sessionIds from the queryString.

src/api/surveillance-form/surveillance-form-keys.ts
```diff
@@ -2,4 +2,6 @@ export const surveillanceFormKeys = {
     root: ['surveillanceForm'] as const,
     surveillanceFormBySessionId: (sessionId: number) =>
         ['surveillanceForm', 'sessionId', sessionId] as const,
+    allSurveillanceForms: (sessionIds: number[]) =>
+        ['surveillanceForm', 'all', sessionIds] as const,
```
aryamanshodhan: Maybe use sessionIds.toSorted() to avoid unnecessary refetches if the arrays are [1, 2, 3] vs. [2, 3, 1]

src/api/surveillance-form/get-all-surveillance-forms.ts
```diff
@@ -0,0 +1,27 @@
...
+} from '@/api/surveillance-form/validation/get-all-surveillance-forms-schema';
+import type { NetworkError } from '@/lib/network/network-error';
+import { ok, type Result } from '@/lib/result/result';
+import { getSurveillanceFormBySessionId } from '@/api/surveillance-form/get-surveillance-form-by-session-id';
+
+export async function getAllSurveillanceForms(
+    accessToken: string,
+    queryParams: GetAllSurveillanceFormsQueryParams,
+): Promise<Result<GetAllSurveillanceFormsResponseBody, NetworkError>> {
+    const results = await Promise.all(
```
aryamanshodhan: Be explicit with variable names. getAllSurveillanceFormsResult

src/api/surveillance-form/validation/get-all-surveillance-forms-schema.ts
```diff
@@ -0,0 +1,21 @@
+import { z } from 'zod';
+import { surveillanceFormSchema } from '@/api/surveillance-form/validation/surveillance-form-schema';
+
+export const getAllSurveillanceFormsQueryParamsSchema = z.object({
+    sessionIds: z.array(z.number()),
```
aryamanshodhan: Rename sessionIds → sessionId and change to z.coerce.number() for query string compatibility. Repeated query params use singular keys by convention (?sessionId=3&sessionId=4), and z.coerce handles string→number parsing from URL params. This aligns with how getSessionsQueryParamsSchema uses z.coerce.number().

sessionId: z.array(z.coerce.number()),

src/api/surveillance-form/get-all-surveillance-forms.ts
```diff
@@ -0,0 +1,27 @@
...
+        queryParams.sessionIds.map(sessionId =>
+            getSurveillanceFormBySessionId(accessToken, sessionId),
+        ),
+    );
+
+    const allSurveillanceForms = results
+        .filter(result => result.ok)
+        .map(result => result.data);
```
aryamanshodhan: I think this is unsafe. If there is a failed result, you probably don;t want to silently skip over it. A safer approach might be something like:

const surveillanceForms: SurveillanceForm[] = [];

for (const result of results) {
    if (!result.ok) return result;
    surveillanceForms.push(result.data);
}

src/features/review/components/site-detail/surveillance-form-review/page-client/surveillance-form-review-page-client.tsx
```diff
@@ -0,0 +1,88 @@
...
+import SurveillanceFormReviewWorkspace from '@/features/review/components/site-detail/surveillance-form-review/surveillance-form-review-workspace';
+import { ClipboardList } from 'lucide-react';
+
+interface SurveillanceFormReviewPageClientProps {
+    siteId: number;
+    startDate?: string;
+    endDate?: string;
+}
+
+export default function SurveillanceFormReviewPageClient({
```
aryamanshodhan: This should be a generic page client for the entire review details page, since all 3 steps are technically happening on the same page. There should be a 1-1 page to page client.

ReviewDetailsPageClient

src/api/surveillance-form/hooks/use-get-all-surveillance-forms.ts
aryamanshodhan, line 1: Switch from POST with JSON.stringify(queryParams) to GET with constructQueryString. This matches useGetAllSessions, useGetSessions, and useGetAnnotations which all use constructQueryString for read operations.

src/app/api/surveillance-forms/all/route.ts
aryamanshodhan, line 1: Switch from POST to GET. This is a read operation — POST implies mutation. Every other read endpoint in the codebase uses GET with query params. Use url.searchParams.getAll('sessionId') to extract the repeated params.

Missing try/catch on body parsing. await request.json() can throw on malformed input. The annotation mutation route (PUT /api/annotations/[id]) wraps this in try/catch — this should too. (Becomes irrelevant with the switch to GET.)

src/api/surveillance-form/get-all-surveillance-forms.ts
aryamanshodhan, line 1: Silent error swallowing. The .filter(result => result.ok) silently drops failed fetches. If 3 out of 5 forms fail, the caller sees 2 forms with a success message and no indication of failure. Every other "get all" function in the codebase (e.g., getAllSessions) propagates the first error immediately. Build the array inside the error-checking loop to avoid both the silent drop and the redundant filter:

const surveillanceForms: SurveillanceForm[] = []; for (const result of results) { if (!result.ok) return result; surveillanceForms.push(result.data); }

Redundant filter. If the error propagation loop passes, every result is guaranteed ok — the .filter(result => result.ok) after is dead code.

src/app/(dashboard)/review/[siteId]/page.tsx
Comment (aryamanshodhan) [inline src/app/(dashboard)/review/[siteId]/page.tsx:1]: Missing server-side prefetch for surveillance forms. Sessions are prefetched and hydrated, but surveillance forms are left to the client — creating a waterfall (page loads → sessions hydrate → client fetches forms). The annotation feature prefetches all needed data server-side. Fetch surveillance forms within the same withAuthSession callback after sessions are available, and set them in the query client for hydration.

Comment (aryamanshodhan) [inline src/app/(dashboard)/review/[siteId]/page.tsx:1]: Route structure: [district] is in the URL path but never used — only [siteId] is destructured and passed to the API. Consider flattening to [/review/[siteId]] for consistency with the annotation feature (/annotate/[annotationTaskId]).

Comment (aryamanshodhan) [inline src/app/(dashboard)/review/[siteId]/page.tsx:1]: Duplicated [endDate] computation: [endOfMonth] + [format] is already computed in the sites list page client. Here it's recomputed from the [month] query param via [parseISO].

Pass both [startDate] and [endDate] as query params from the site card link, so this page just reads them.

Comment (aryamanshodhan) [inline src/app/(dashboard)/review/[siteId]/page.tsx:1]: [month] param naming is misleading: The value is "2026-03-01" (a full date), not "2026-03" (a month). It's used as [startDate] directly:

const startDate = month;

src/features/review/components/site-detail/surveillance-form-review/types.ts
aryamanshodhan, line 1: I think just inline this into whichever file you are using

src/features/review/components/site-detail/surveillance-form-review/page-client/surveillance-form-review-page-client.tsx
aryamanshodhan, line 1: Update { sessionIds } → { sessionId: sessionIds } to match the renamed schema field.

aryamanshodhan, line 1: Data fetching and joining should move to the workspace component. The page client fetches sessions, derives sessionIds, fetches surveillance forms, builds a formsMap, and joins them — that's all data composition logic. In the annotation feature, the page client manages filter state and layout, while the workspace component owns data fetching and transformation. Move the useGetSessions, useGetAllSurveillanceForms, and the formsMap joining into SurveillanceFormReviewWorkspace. The page client should just pass siteId, startDate, endDate down.

aryamanshodhan, line 1: With server-side prefetching, loading/error states simplify. If the server page prefetches both sessions and surveillance forms (and redirects on auth errors), the client component will have hydrated data on first render. The loading/error early returns become fallbacks for client-side refetches rather than the primary code path.

aryamanshodhan, line 1: sessionIds is re-derived on every render. [sessions.map(s => s.sessionId)] creates a new array reference each render. Consider useMemo for stability:

`const sessionIds = useMemo(
    () => sessions.map(s => s.sessionId),
    [sessions],
);`

aryamanshodhan, line 1: Description string uses a template literal with no interpolation. description={`Review submitted session data by district and month.`} — the backticks can be replaced with regular quotes since there's no dynamic content.

src/features/review/components/site-detail/surveillance-form-review/surveillance-form-review-table.tsx
aryamanshodhan, line 1: Simplify DATA_FIELDS — remove the sessionKey shortcut and unify on a single pattern. The current approach has two code paths: sessionKey (direct property access) and render (custom function), which creates branching logic in the table body. Use a consistent render function for every field. This eliminates the if/else in the cell rendering and makes every field follow the same pattern.

aryamanshodhan, line 1: Consider a shared formatValue utility for null/boolean formatting. Multiple render functions repeat the same form ? ... : 'N/A' and value ? 'Yes' : 'No' patterns. A generic formatter would reduce duplication:

function formatValue(value: unknown): string {
    if (value === null || value === undefined) return 'N/A';
    if (typeof value === 'boolean') return value ? 'Yes' : 'No';
    return String(value);
}

src/features/review/components/site-detail/review-site-detail-header.tsx
aryamanshodhan, line 1: The connector line renders after the last step's content div but inside the same gap-4 container. This means the last step gets a trailing gap with no connector. Not a visual bug because of the index < Steps.length - 1 guard, but the outer gap-4 on the step wrapper still applies spacing after the connector. Consider moving the connector outside the step's div or using CSS gap only on the parent.

aryamanshodhan, line 1: Pass Steps as a prop to ReviewSiteDetailHeader instead of hardcoding it inside. Define the steps array one layer above (in the page client) and pass it down. This makes the header a reusable stepper component with no hardcoded knowledge of the review workflow steps.

src/features/review/components/sites-list/review-site-card.tsx
```diff
@@ -10,6 +10,7 @@ interface ReviewSiteCardProps {
     site: Site;
     sessionCount: number;
     state?: SessionState;
+    month: string;
```
aryamanshodhan: reconsider nomenclature

src/features/review/components/sites-list/review-sites-list.tsx
```diff
@@ -101,6 +101,7 @@ export default function ReviewSitesList({
                             site={site}
                             sessionCount={count ?? 0}
                             state={state}
+                            month={startDate}
```
aryamanshodhan: reconsider month nomenclature

aryamanshodhan (changes requested):

src/api/surveillance-form/hooks/use-get-all-surveillance-forms.ts
aryamanshodhan, line 1: Question: Instead of this endpoint, should we use the GET all sessions endpoint with the getSurveillanceFormForSessionId endpoint?

src/api/surveillance-form/validation/get-all-surveillance-forms-schema.ts
```diff
@@ -0,0 +1,21 @@
+import { z } from 'zod';
+import { surveillanceFormSchema } from '@/api/surveillance-form/validation/surveillance-form-schema';
+
+export const getAllSurveillanceFormsQueryParamsSchema = z.object({
+    sessionId: z.array(z.coerce.number()),
+});
```
aryamanshodhan: This would be the same as just calling the get surveillance form by session Id endpoint. Just trying to reduce the total number of files in the codebase, especially because there are the same number of API calls.

src/api/surveillance-form/get-all-surveillance-forms.ts
aryamanshodhan, line 1: Check comments above. I think we have opportunity to simplify and remove all these files.

src/api/surveillance-form/surveillance-form-keys.ts
aryamanshodhan, line 1: See above. opportunity to remove these changes as well.

src/app/(dashboard)/review/[siteId]/page.tsx
Comment (aryamanshodhan) [inline src/app/(dashboard)/review/[siteId]/page.tsx:1]: [district] does not account for generic siteIds and non-Uganda programs. Since siteIds are unique, you can remove the [district] dynamic page. Just have [siteId]

Comment (aryamanshodhan) [inline src/app/(dashboard)/review/[siteId]/page.tsx:1]: Question: Is this a separate page that opens up? Or is it on the same page?

src/app/api/surveillance-forms/all/route.ts
aryamanshodhan, line 1: See comments above. Opportunity for this file to go away as well.

src/features/review/components/sites-list/review-site-leaf-rows.tsx
```diff
@@ -25,24 +33,31 @@ export default function ReviewSiteLeafRows({
...
-                    >
+                const rowClassName = cn(
+                    'flex items-center justify-between rounded-md px-3 py-2',
+                    hasSessions
+                        ? 'cursor-pointer hover:bg-muted/50'
+                        : 'cursor-not-allowed opacity-60',
+                );
+
+                const rowContent = (
+                    <>
```
aryamanshodhan: Use Fragment. More explcit

src/features/review/components/sites-list/review-site-leaf-rows.tsx
```diff
@@ -25,24 +33,31 @@ export default function ReviewSiteLeafRows({
                         needsReviewCount: 0,
                     };
                 const hasSessions = sessionCount > 0;
+                const href = `/review/${site.district}/${site.siteId}?startDate=${startDate}&endDate=${endDate}`;
```
aryamanshodhan: Remove district from the href

src/features/review/components/sites-list/review-sites-list.tsx
```diff
@@ -159,6 +159,7 @@ export default function ReviewSitesList({
                                 sites={sites}
                                 depth={0}
                                 parentPath={monthKey}
+                                monthKey={monthKey}
```
aryamanshodhan: Question: explain the need for monthKey. Is there opportunity to derive a unique key inside the component itself? There is multiple layers of prop drilling happening, especially when parentPath is also just the monthKey.

src/features/review/components/site-detail/surveillance-form-review/surveillance-form-review-table.tsx
```diff
@@ -0,0 +1,120 @@
...
+    Table,
+    TableBody,
+    TableCell,
+    TableHead,
+    TableHeader,
+    TableRow,
+} from '@/components/ui/table';
+import { format } from 'date-fns';
+
+interface SessionWithForm {
```
aryamanshodhan: I would define separate schema files for composite domain models

src/features/review/components/site-detail/surveillance-form-review/surveillance-form-review-table.tsx
```diff
@@ -0,0 +1,120 @@
...
+        label: 'LLINs Available',
+        render: (_, form) => formatValue(form?.numLlinsAvailable),
+    },
+    { label: 'LLIN Type', render: (_, form) => formatValue(form?.llinType) },
+    { label: 'LLIN Brand', render: (_, form) => formatValue(form?.llinBrand) },
+    {
+        label: 'People Under LLIN',
+        render: (_, form) => formatValue(form?.numPeopleSleptUnderLlin),
+    },
+];
```
aryamanshodhan, line 52: What about dynamic form questions and answers? If you don't know what I am talking about, text me. This will not work with Ghana because they have a different form schema.

src/features/review/components/site-detail/surveillance-form-review/surveillance-form-review-workspace.tsx
```diff
@@ -0,0 +1,107 @@
+'use client';
+
+import { useGetSessions } from '@/api/session/hooks/use-get-sessions';
+import { useGetAllSurveillanceForms } from '@/api/surveillance-form/hooks/use-get-all-surveillance-forms';
+import type { Session } from '@/api/session/validation/session-schema';
+import type { SurveillanceForm } from '@/api/surveillance-form/validation/surveillance-form-schema';
+import SurveillanceFormReviewTable from '@/features/review/components/site-detail/surveillance-form-review/surveillance-form-review-table';
+import { useMemo } from 'react';
+
+interface SessionWithForm {
```
aryamanshodhan: Same as previous. Composite domain models should get their own schema file

src/features/review/components/site-detail/surveillance-form-review/surveillance-form-review-workspace.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+export default function SurveillanceFormReviewWorkspace({
+    siteId,
+    startDate,
+    endDate,
+}: SurveillanceFormReviewWorkspaceProps) {
+    const { data: getSessionsResult, isPending: isGetSessionsPending } =
+        useGetSessions({ siteId, startDate, endDate });
+
+    const { sessions, sessionIds } = useMemo(() => {
```
aryamanshodhan: I think you can just use sessions here and derive sessionIds inline.

src/features/review/components/site-detail/review-site-detail-header.tsx
aryamanshodhan, line 1: Question: This is for step 1 right?

src/features/review/components/site-detail/review-site-detail-header.tsx
```diff
@@ -0,0 +1,73 @@
+'use client';
+
+import { Check } from 'lucide-react';
+import { Fragment } from 'react';
+import { cn } from '@/utils/cn';
+
+interface Step {
```
aryamanshodhan: I think this is a redundant interface.

aryamanshodhan (changes requested):
Great start Cole! There are a few changes needed. Please submit an updated PR.

src/api/surveillance-form/validation/get-surveillance-form-by-session-id-schema.ts
```diff
@@ -8,5 +12,8 @@ export type GetSurveillanceFormBySessionIdResponseBody = z.infer<
     typeof getSurveillanceFormBySessionIdResponseSchema
 >;
 
-export type GetSurveillanceFormBySessionIdSuccessPayload =
-    GetSurveillanceFormBySessionIdResponseBody;
+export type SurveillanceFormData =
```
aryamanshodhan, line 15: I am guessing we will remove this as soon as we phase out the legacy structure?

src/api/surveillance-form/validation/session-with-rows-schema.ts
```diff
@@ -0,0 +1,15 @@
+import { sessionSchema } from '@/api/session/validation/session-schema';
+import { z } from 'zod';
+
+const formRowSchema = z.object({
+    label: z.string(),
+    value: z.string(),
+});
+
+export const sessionWithRowsSchema = z.object({
```
aryamanshodhan: Question: explain this schema.

src/api/surveillance-form/get-surveillance-form-data-by-session-id.ts
```diff
@@ -0,0 +1,32 @@
+import { ok, type Result } from '@/lib/result/result';
+import type { NetworkError } from '@/lib/network/network-error';
+import { getFormAnswersBySessionId } from '@/api/surveillance-form/get-form-answers-by-session-id';
+import { getSurveillanceFormBySessionId } from '@/api/surveillance-form/get-surveillance-form-by-session-id';
+import type { SurveillanceFormData } from '@/api/surveillance-form/validation/get-surveillance-form-by-session-id-schema';
+
+export async function getSurveillanceFormDataBySessionId(
+    accessToken: string,
+    sessionId: number,
+    version?: string,
```
aryamanshodhan, line 10: Just letting you know in case anything can be simplified: 

GET /forms/current will return the latest form and form questions.

src/features/review/components/site-detail/surveillance-form-review/surveillance-form-review-workspace.tsx
```diff
@@ -64,27 +91,37 @@ export default function SurveillanceFormReviewWorkspace({
...
+        if (query.data?.ok) {
+            formsMap.set(
+                query.data.data.sessionId,
+                normalizeFormData(query.data.data),
+            );
+        }
+    }
 
-    const surveillanceForms: SessionWithForm[] = sessions.map(session => ({
+    const surveillanceForms: SessionWithRows[] = sessions.map(session => ({
```
aryamanshodhan: What is SessionWithRows? Rows of what?

src/features/review/components/sites-list/review-site-leaf-rows.tsx
```diff
@@ -18,10 +19,10 @@ interface ReviewSiteLeafRowsProps {
 export default function ReviewSiteLeafRows({
     sites,
     getDisplayName,
-    monthKey,
+    parentPath,
     sessionCountsBySiteId,
 }: ReviewSiteLeafRowsProps) {
-    const startDate = `${monthKey}-01`;
+    const startDate = `${parentPath.split('/')[0]}-01`;
```
aryamanshodhan: Isn't this fragile? I wonder if there is another way to handle this.

aryamanshodhan (commented):
Minor comments. Code mostly is ready.

src/api/surveillance-form/validation/get-surveillance-form-by-session-id-schema.ts
csmanel: I think so, unless you have another suggestion

src/api/surveillance-form/validation/session-with-rows-schema.ts
csmanel: This pairs a session with its surveillance form data formatted into label/value rows for the review table `{ label: "People in House", value: "5" }`. 
rows are null when no form was submitted for that session.

src/api/surveillance-form/get-surveillance-form-data-by-session-id.ts
csmanel: I don't think this will help simplify anything at the moment. But maybe you see something I am not

src/features/review/components/site-detail/surveillance-form-review/surveillance-form-review-workspace.tsx
csmanel: Renaming it to `SessionWithFormFieldRows`
They are `{ label: string, value: string }` pairs derived from form answers/legacy surveillance form fields. 
They are just label/value pairs for the table.

src/app/(dashboard)/review/[siteId]/page.tsx
Comment (csmanel) [reply src/app/(dashboard)/review/[siteId]/page.tsx:1]: Yes it is a new page

aryamanshodhan (approved):

csmanel (approved):

PR #83, opened 2026-04-01
VCV-148: Dark Mode for the UI

csmanel:
Add dark mode setting

### DIFF
--- a/src/app/layout.tsx
+++ b/src/app/layout.tsx
@@ -1,5 +1,6 @@
 import '@/app/globals.css';
 import { TanstackProvider } from '@/components/providers/tanstack-provider';
+import { ThemeProvider } from '@/components/providers/theme-provider';
 import { TooltipProvider } from '@/components/ui/tooltip';
 import type { Metadata } from 'next';
 import { Geist, Geist_Mono } from 'next/font/google';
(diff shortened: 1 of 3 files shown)


src/app/layout.tsx
```diff
@@ -26,13 +27,20 @@ export default function RootLayout({
     children: React.ReactNode;
 }>) {
     return (
-        <html lang="en" className="h-full">
+        <html lang="en" className="h-full" suppressHydrationWarning>
```
aryamanshodhan, line 30: Question: why are we suppressing warning?

src/components/layout/app-sidebar.tsx
aryamanshodhan: Lets move this hook up with the other hooks.

src/components/providers/theme-provider.tsx
```diff
@@ -0,0 +1,11 @@
+'use client';
```
aryamanshodhan: Question: Why does this need to be a client component?

aryamanshodhan (commented):
Looks good Cole. Just a couple of very minor questions.

src/app/layout.tsx
csmanel: [docs](https://github.com/pacocoursey/next-themes)
"Note! If you do not add [suppressHydrationWarning](https://reactjs.org/docs/dom-elements.html#suppresshydrationwarning:~:text=It%20only%20works%20one%20level%20deep) to your <html> you will get warnings because next-themes updates that element. This property only applies one level deep, so it won't block hydration warnings on other elements."

src/components/providers/theme-provider.tsx
csmanel: it doesn't :)

aryamanshodhan (approved):

PR #86, opened 2026-04-09
VCV-164: Create Operations Overview page 

csmanel:
- Creates skeleton props for consistent display (values may need to change as we use them in more places) 
- Display Skeleton on sites page before a district has been selected, when districts are loading, or when there is an error loading districts


src/components/ui/skeleton.tsx
aryamanshodhan, line 1: Just confirming, this is changing the shadcn skeleton correct? 

I will urge you to think about whether there is an alternative way in which we can define custom skeleton components which does not involve needing to pass in 4-5 props. Isn't there a way to do this using tailwind css classes?

src/features/operations/components/page-client/operations-page-client.tsx
```diff
@@ -100,11 +101,23 @@ export default function OperationsPageClient() {
                     <Separator />
 
                     {!selectedDistrict ? (
-                        <div className="flex flex-col items-center justify-center py-12 text-center">
-                            <p className="text-muted-foreground text-sm">
-                                Select a district to view data.
-                            </p>
-                            <Microscope className="text-muted-foreground/50 mt-4 h-12 w-12" />
+                        <div className="relative">
+                            <div className="space-y-1">
```
aryamanshodhan: Question: Should we move the skeleton for this page to a different component file?

src/features/operations/components/site-list/operations-site-list.tsx
```diff
@@ -50,12 +51,40 @@ export default function OperationsSiteList({
         });
     }
 
+    const skeletonCount = new Set(sites.map(s => s.subCounty)).size || 5;
+
     if (isGetAllSessionsPending || !getAllSessionsResult) {
-        return <h1>LOADING...</h1>;
+        return (
```
aryamanshodhan: Question: Should we define a new loading skeleton component for this?

src/features/operations/components/site-list/operations-site-list.tsx
```diff
@@ -50,12 +51,40 @@ export default function OperationsSiteList({
...
+                        <Skeleton height="xl" width="full" />
+                    </div>
+                ))}
+            </div>
+        );
     }
 
     if (!getAllSessionsResult.ok) {
-        return <h1>ERROR: {getAllSessionsResult.error.message}</h1>;
+        return (
```
aryamanshodhan, line 99: Same here

aryamanshodhan (commented):
Hey Cole! Minor changes recommended for this PR. Please look through it. Thanks!

src/components/ui/skeleton.tsx
csmanel: Yes, this modifies the shadcn skeleton. We could leave it alone and create a fully custom component instead, happy to do that if preferred.

The props approach was intentional, my hope is that it enforces consistent sizing tokens so skeletons don't end up with arbitrary one-off values scattered across the codebase. The idea is that common loading scenarios share the same sizes rather than each defining their own. I would rather provide a size prop, than copy the same tailwind class across each skeleton.

src/features/operations/components/page-client/operations-page-client.tsx
csmanel: It is your call. In this moment I think if it gets used anywhere else we absolutely should, but at the moment its only being used here and I don't know if we gain much from that abstraction?

aryamanshodhan (dismissed):
Looks good Cole! Please do a final round of testing and merge this in.

src/features/operations/components/site-list/operations-site-list.tsx
```diff
@@ -2,69 +2,127 @@
...
+        });
     }
 
-    if (!getAllSessionsResult.ok) {
+    const skeletonCount = new Set(sites.map(site => site.subCounty)).size || 5;
+
+    if (
+        isGetAllSessionsPending ||
+        !getAllSessionsResult ||
+        !getAllSessionsResult.ok
```
aryamanshodhan: If !getAllSessionsResult.ok

This should be a separate failure branch displaying an error. Otherwise, the UI will perpetually be in a loading state

aryamanshodhan (changes requested):

aryamanshodhan (approved):

PR #92, opened 2026-04-18
VCV-143: Calculate and display coverage percentage and other annotation metrics within the AI performance tab

jyoon53:
- Implemented real coverage and annotation metrics in the AI Performance tab, replacing the static placeholder UI                                                                     
- Added a full three-layer API integration for GET /annotations/summary (schema → server fetcher → BFF route → React Query hook)                                                      
-  Hardened the specimens/count Zod schema to prevent silent validation failures caused by nullable fields (species, sex, abdomenStatus, siteInfo) that weren't marked .nullable() —   
- safeApiCall was silently returning errors, causing all metrics to display —                                                                                                           
- Merged the Coverage and Validated Specimens metrics into a single card per design intent                                                                                            
- Sourced Last Update from GET /annotations/task, filtered to tasks with at least one annotated specimen, and resolved to the max updatedAt timestamp                                 

Known Limitation: Last Update is not district-scoped                                                                                                                                 
-   GET /annotations/task does not support district filtering — it returns annotation tasks program-wide. This means when querying Last Update for a specific district, tasks from other districts are included in the result set.
- Symptom: A district with 0% coverage (no annotated specimens) can show a non-null Last Update timestamp sourced from another district's activity — which is logically incorrect.      
- Temporary fix: Added a validatedSpecimens > 0 guard before computing the Last Update timestamp. If the selected district has no annotated specimens, Last Update displays — regardless of what the tasks endpoint returns.


jyoon53:
merged into 144

PR #105, opened 2026-04-20
hide the ai performance tab from users that dont have permission to r…

csmanel:
…ead and write annotation tasks

### DIFF
--- a/src/features/operations/components/page-client/operations-page-client.tsx
+++ b/src/features/operations/components/page-client/operations-page-client.tsx
@@ -15,12 +15,26 @@ import { SkeletonList } from '@/components/ui/skeleton-list';
 import ExportDialog from '@/features/operations/components/export/export-dialog';
 import OperationsSpeciesComposition from '../species-composition/operations-species-composition';
 import { useLocationSelection } from '../../hooks/use-location-selection';
+import type { UserPermissions } from '@/api/user/validation/user-permissions-schema';
 
 const OPERATIONS_TABS = [
-    { value: 'sites', label: 'SITES' },
-    { value: 'geographical-summary', label: 'GEOGRAPHICAL SUMMARY' },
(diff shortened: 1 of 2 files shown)


src/features/operations/components/page-client/operations-page-client.tsx
```diff
@@ -15,12 +15,26 @@ import { SkeletonList } from '@/components/ui/skeleton-list';
...
 const OPERATIONS_TABS = [
-    { value: 'sites', label: 'SITES' },
-    { value: 'geographical-summary', label: 'GEOGRAPHICAL SUMMARY' },
-    { value: 'species-composition', label: 'SPECIES COMPOSITION' },
-    { value: 'ai-performance', label: 'AI PERFORMANCE' },
+    { value: 'sites', label: 'SITES', shouldRender: () => true },
+    {
+        value: 'geographical-summary',
+        label: 'GEOGRAPHICAL SUMMARY',
+        shouldRender: () => true,
```
aryamanshodhan, line 25: question: Will this cause bug if this has no argument but the other has?

aryamanshodhan (approved):

src/features/operations/components/page-client/operations-page-client.tsx
csmanel: I believe it will be ok, but if you want we can be explicit and have them each as 
```permissions: UserPermissions) => boolean```

edit: it looks like the linter would be upset if we gave it `permissions` that were not used

aryamanshodhan: I see. But does this cause a compilation error? because youre saying some take in no arguments, others do. And then you are passing permissions into all of them anyway. I know you can replace unused parameters with _ in functions. So you can have something like:

shouldRender: (_permissions : UserPermissions) => true

aryamanshodhan (approved):

PR #106, opened 2026-04-20
hide the ai performance tab from users that dont have permission to r…

csmanel:
…… (#105)

* hide the ai performance tab from users that dont have permission to read and write annotation tasks

* prettier

* add should render filtering logic on opreations tabs

* add country to signup form alongside program name


aryamanshodhan (dismissed):
Remember, MERGE COMMIT, NOT squash and merge

aryamanshodhan (approved):

PR #124, opened 2026-05-22
VCV-204: Local storage Operations and Review

csmanel:
## Summary

Persists UI state in the Review and Operations pages so users don't lose their
place when navigating away or switching tabs.

- **`useLocalStorage` hook** : SSR-safe, Date-aware, silently falls back on
  corrupt data
- **`StorageKeys`** : centralized constant object, no inline key literals
- **Review & Operations**: `activeTab`, `startMonth`, `endMonth`, and
  `selectedLocation` now survive page navigation via localStorage
- **Review**: `expandedSitePaths` and `collapsedMonths` lifted to page client
  so expanded rows survive switching to the Export tab
- **Operations**: `selectedMarkerId` lifted to page client so the active map
  marker survives tab switches; resets when filters change


src/features/operations/geographical-summary/components/site-map.tsx
```diff
@@ -84,7 +87,7 @@ export default function SiteMap({ markers, selectedLocation }: SiteMapProps) {
             <MarkerInfoPanel
                 markers={markers}
                 selectedMarkerId={selectedMarkerId}
-                onMarkerSelect={setSelectedMarkerId}
+                onMarkerSelect={onMarkerSelect}
```
aryamanshodhan, line 90: What is onMarkerSelect doing? I dont see it being used as part of this PR, seems like it is only prop drilling.

src/features/operations/geographical-summary/components/operations-geographical-summary.tsx
```diff
@@ -78,6 +84,8 @@ export default function OperationsGeographicalSummary({
                         <SiteMap
                             markers={markers}
                             selectedLocation={selectedLocation}
+                            selectedMarkerId={selectedMarkerId}
```
aryamanshodhan, line 87: Seems like this is just prop drilling. How is this being used in the SiteMap? Seems like this was added this PR, and there is no other way I see this is being used in the map.

src/features/review/sites-list/components/sites/review-sites-list.tsx
```diff
@@ -22,21 +22,22 @@ interface ReviewSiteListProps {
...
 
 export default function ReviewSitesList({
     sites,
     locationQueryParam,
     startMonth,
     endMonth,
+    expandedSitePaths,
+    setExpandedSitePaths,
+    collapsedMonths,
+    setCollapsedMonths,
```
aryamanshodhan, line 43: Lots of props in this function. Why did we remove local state and make them props instead?

src/features/review/sites-list/components/review-sites-list-page-client.tsx
```diff
@@ -61,7 +70,22 @@ export default function ReviewSitesListPageClient() {
...
+        accessibleSites,
+        StorageKeys.review.selectedLocation,
+    );
+
+    const [expandedSitePaths, setExpandedSitePaths] = useState<Set<string>>(
+        new Set(),
+    );
+    const [collapsedMonths, setCollapsedMonths] = useState<Set<string>>(
+        new Set(),
+    );
```
aryamanshodhan, line 84: Why not keep this inside the component that consumes this?

src/lib/storage/hooks/use-local-storage.ts
aryamanshodhan, line 1: This is a utility hook that can be used throughout the project. I would move it out of this specific feature to the root hooks folder (same place where usePagination exists)

src/lib/storage-keys.ts
aryamanshodhan, line 1: Same as above. move out of feature specific folder

aryamanshodhan (changes requested):
Looks great Cole! Just some minor changes.

src/features/operations/components/operations-page-client.tsx
```diff
@@ -44,13 +49,20 @@ const OPERATIONS_TABS = [
...
-    const [activeTab, setActiveTab] = useState<OperationsTab>(
+    const [activeTab, setActiveTab] = useLocalStorage<OperationsTab>(
+        StorageKeys.operations.activeTab,
         'geographical-summary',
     );
-    const [startMonth, setStartMonth] = useState(() =>
+    const [startMonth, setStartMonth] = useLocalStorage(
+        StorageKeys.operations.startMonth,
         startOfMonth(subMonths(new Date(), 2)),
+        DATE_SERIALIZER,
```
aryamanshodhan: Reconsider this. Is there a world where we can have the serialization happen within the hook by just checking types?

src/lib/hooks/use-local-storage.ts
```diff
@@ -0,0 +1,70 @@
...
+export const SET_SERIALIZER: LocalStorageOptions<Set<string>> = {
+    serialize: (value: Set<string>) => JSON.stringify(Array.from(value)),
+    deserialize: (serializedValue: string) =>
+        new Set<string>(JSON.parse(serializedValue)),
+};
+
+export function useLocalStorage<T>(
+    key: string,
+    defaultValue: T,
+    options?: LocalStorageOptions<T>,
```
aryamanshodhan, line 41: If there a way we can check for types? If so, maybe we do not need to pass in serializers that are already defined in the same file. It feels strange importing the serializer from this file, only to re-feed it into useLocalStorage also in this file :)

src/lib/storage-keys.ts
```diff
@@ -0,0 +1,17 @@
...
+        expandedSitePaths: 'review.expandedSitePaths',
+        collapsedMonths: 'review.collapsedMonths',
+    },
+    operations: {
+        activeTab: 'operations.activeTab',
+        startMonth: 'operations.startMonth',
+        endMonth: 'operations.endMonth',
+        selectedLocation: 'operations.selectedLocation',
+        selectedMarkerId: 'operations.selectedMarkerId',
+    },
```
aryamanshodhan, line 16: Question: Do we need anything for annotation?

aryamanshodhan (changes requested):

src/lib/storage-keys.ts
csmanel: This task was just for operations and review

aryamanshodhan (approved):

PR #126, opened 2026-05-26
VCV-212: Remove UNKNOWN and Cannot be Determined labels from being considered in the sensitivity/specificity calculations because it misrepresents accuracy

scho324:
Removed UNKNOWN and Cannot be Determined labels from being considered in the sensitivity/specificity calculations because it misrepresents accuracy. 
* Filtered labels before the calculations to exclude them
* Added number of UNKNOWN/Cannot be Determined specimens below the table
* Updated the interpretation to note the removal of those labels and the meaning of an UNKNOWN label


csmanel (commented):
This is a great start! 
I think there are a few things I think we may need to look at before merging.

src/features/operations/ai-performance/components/specimen-confusion-matrix.tsx
```diff
@@ -91,6 +91,13 @@ export default function SpecimenConfusionMatrix({
             ? totalCorrectPredictions / totalSpecimensInMatrix
             : null;
 
+    const unknownSpecimens = classLabels.filter(
+        label => label === 'UNKNOWN',
+    ).length;
+    const cannotBeDeterminedSpecimens = classLabels.filter(
```
csmanel: I believe you can use these helpers to get the actual count per specimen 

```
const unknownSpecimens = getSpecimenCountForGroundTruthLabel('UNKNOWN');

const cannotBeDeterminedSpecimens = getSpecimenCountForGroundTruthLabel('Cannot be Determined');
```

As it stands, I think this will only ever give us either 0 or 1 for the count, rather than displaying the actual specimen count.

src/features/operations/ai-performance/components/specimen-confusion-matrix.tsx
```diff
@@ -229,60 +236,74 @@ export default function SpecimenConfusionMatrix({
...
+                                        const falseNegatives =
+                                            getSpecimenCountForGroundTruthLabel(
+                                                classLabel,
+                                            ) - truePositives;
+                                        const falsePositives =
+                                            getSpecimenCountForPredictedLabel(
+                                                classLabel,
+                                            ) - truePositives;
+                                        const trueNegatives =
+                                            totalSpecimensInMatrix -
```
csmanel: I am curious, are FN and FP being influenced by UNKNOWN here? Are the numbers being fully filtered as expected?

aryamanshodhan: Yes, they are being influenced! That is exactly what we should not be doing lol because it misrepresents the accuracy. We should exclude those numbers from the calcualtion of sensitivity and specificity.

aryamanshodhan (changes requested):

src/features/operations/ai-performance/components/specimen-confusion-matrix.tsx
scho324: Hopefully it works now, I adjusted getSpecimenCountFor(GroundTruth/Predicted)Label to take a labels array (classLabels by default so existing uses don't have to be changed) so the calls to them for FN and FP use the filtered labels array and not the original one

src/features/operations/ai-performance/components/specimen-confusion-matrix.tsx
```diff
@@ -283,6 +313,11 @@ export default function SpecimenConfusionMatrix({
                                 })}
                             </TableBody>
                         </Table>
+                        <p className="text-muted-foreground text-sm leading-6">
+                            {cannotBeDeterminedSpecimens} specimens were labeled
```
csmanel: Avoid hardcoded user-facing text; use localized strings from our i18n system instead.
There is a doc for how that works here `.claude/docs/i18n.md`

csmanel (commented):
This looks great! One minor change.

src/features/operations/ai-performance/components/specimen-confusion-matrix.tsx
```diff
@@ -91,6 +100,28 @@ export default function SpecimenConfusionMatrix({
             ? totalCorrectPredictions / totalSpecimensInMatrix
             : null;
 
+    const unknownSpecimens = getSpecimenCountForGroundTruthLabel('UNKNOWN');
+    const cannotBeDeterminedSpecimens = getSpecimenCountForGroundTruthLabel(
+        'Cannot be Determined',
```
csmanel: If we use `UNKNOWN` and `cannot be determined` enough times should we just export them to their own const? 

something like 

`const EXCLUDED_LABELS = ['UNKNOWN', 'Cannot be Determined'] as const;`

src/features/operations/ai-performance/components/specimen-confusion-matrix.tsx
```diff
@@ -283,34 +318,35 @@ export default function SpecimenConfusionMatrix({
...
-                            Interpretation
+                            {t('interpretation')}
                         </h3>
                         <p className="text-muted-foreground text-sm leading-6">
-                            The matrix compares visual verification{' '}
-                            {classificationCategory} labels from{' '}
-                            {selectedLocationName} against the VectorCam{' '}
-                            {classificationCategory} prediction for the selected
-                            date range.
+                            {t('theMatrixCompares')} {classificationCategory}{' '}
```
csmanel: This is close, but we can utilize ICU interpolation here. 

For example, 
`"matrixCompares": "The matrix compares visual verification {category} labels from {location} against the VectorCam {category} prediction for the selected date range."`

`{t('matrixCompares', { category: classificationCategory, location: selectedLocationName })}`

There are few places where we can make this adjustment, so comb through the file one more time then we should be good to go.

csmanel (changes requested):

src/features/operations/ai-performance/components/specimen-confusion-matrix.tsx
```diff
@@ -91,6 +100,28 @@ export default function SpecimenConfusionMatrix({
             ? totalCorrectPredictions / totalSpecimensInMatrix
             : null;
 
+    const EXCLUDED_LABELS = ['UNKNOWN', 'Cannot be Determined'] as const;
```
csmanel: This should be defined at the module level

csmanel (dismissed):

src/features/operations/ai-performance/components/specimen-confusion-matrix.tsx
```diff
@@ -91,6 +102,27 @@ export default function SpecimenConfusionMatrix({
             ? totalCorrectPredictions / totalSpecimensInMatrix
             : null;
 
+    const excludedSpecimenCounts = EXCLUDED_LABELS.map(label =>
+        getSpecimenCountForGroundTruthLabel(label),
+    );
+
+    const filteredClassLabels = classLabels.filter(
+        label => !(EXCLUDED_LABELS as readonly string[]).includes(label),
```
aryamanshodhan: no need to cast this to readonly string[]

src/features/operations/ai-performance/components/specimen-confusion-matrix.tsx
```diff
@@ -25,6 +26,8 @@ interface SpecimenConfusionMatrixProps {
     selectedLocationName: string;
 }
 
+const EXCLUDED_LABELS = ['UNKNOWN', 'Cannot be Determined'] as const;
```
aryamanshodhan: Move this above the interface SpecimenConfusionMatrixProps
change to ['unknown', 'cannot be determined']

When checking this array, make sure you cast all strings to lowercase before comparing to this array. That way we exclude UNKNOWN, unknown, and Unknown from the calculations.

src/features/operations/ai-performance/components/specimen-confusion-matrix.tsx
```diff
@@ -55,17 +60,23 @@ export default function SpecimenConfusionMatrix({
             : 0;
     }
 
-    function getSpecimenCountForGroundTruthLabel(groundTruthLabel: string) {
-        return classLabels.reduce(
+    function getSpecimenCountForGroundTruthLabel(
+        groundTruthLabel: string,
+        labels = classLabels,
```
aryamanshodhan: Rename to allLabels

src/features/operations/ai-performance/components/specimen-confusion-matrix.tsx
```diff
@@ -283,34 +324,37 @@ export default function SpecimenConfusionMatrix({
...
                             </TableBody>
                         </Table>
+                        <p className="text-muted-foreground text-sm leading-6">
+                            {EXCLUDED_LABELS.map((label, index) => {
+                                return t('specimensLabeledAs', {
+                                    num: excludedSpecimenCounts[index] ?? 0,
+                                    category: label,
+                                });
+                            })}
+                        </p>
```
aryamanshodhan, line 347: What is this block of code doing?

aryamanshodhan (changes requested):

src/features/operations/ai-performance/components/specimen-confusion-matrix.tsx
scho324: TypeScript will say that argument of type 'string' is not assignable to parameter of type '"unknown" | "cannot be determined"' otherwise

src/features/operations/ai-performance/components/specimen-confusion-matrix.tsx
scho324: Displays '{num} specimens were labeled as {category}' based on what labels are in EXCLUDED_LABELS and the counts for those labels in excludedSpecimenCounts (Jira ticket said 'Add a small footnote mentioning how many specimens could not be determined or had unknown.')
The ?? 0 part was because TypeScript thought excludedSpecimenCounts[index] was of type number | undefined.

aryamanshodhan: Got it!

src/features/operations/ai-performance/components/specimen-confusion-matrix.tsx
aryamanshodhan: The labels should then be typed according to the annotation schema. Check out the list of possible values they can choose from the schema, and cast to that type. using a string list is way too generic when you only have a select list of values the user can choose from.

aryamanshodhan (approved):
Well done Sophia! The code looks good!

PR #132, opened 2026-05-28
VCV-199: Create universal collection cycle

csmanel:


src/api/collection-cycle/validation/collection-cycle-schema.ts
aryamanshodhan, line 1: tbh I would separate schedules from cycles from a folder structure perspective

src/api/collection-cycle/validation/collection-cycle-schema.ts
```diff
@@ -0,0 +1,45 @@
...
+    endDate: z.string(),
+});
+
+export const getCollectionCyclesResponseSchema = z.object({
+    collectionCycles: z.array(collectionCycleSchema),
+});
+
+export type CollectionSchedule = z.infer<typeof collectionScheduleSchema>;
+export type CollectionCycle = z.infer<typeof collectionCycleSchema>;
+export type GetCollectionCyclesQueryParams = z.infer<
```
aryamanshodhan: define a separate schema file for getting collection cycles, even though it seems like a trivial file.

src/api/collection-cycle/get-collection-cycles.ts
```diff
@@ -0,0 +1,31 @@
...
+} from '@/api/collection-cycle/validation/collection-cycle-schema';
+import type { Result } from '@/lib/result/result';
+import { constructQueryString } from '@/lib/network/construct-query-string';
+import { safeApiCall } from '@/lib/network/safe-api-call';
+
+export async function getCollectionCycles(
+    accessToken: string,
+    queryParams: GetCollectionCyclesQueryParams,
+): Promise<Result<GetCollectionCyclesResponseBody, NetworkError>> {
+    const queryString = constructQueryString(
```
aryamanshodhan: `constructQueryString<GetCollectionCyclesQueryParams>{ ... }`

src/features/review/sites-list/components/layout/review-sites-list-header.tsx
aryamanshodhan, line 1: Shouldn't the collection cycle picker be rendered here?

src/features/review/sites-list/components/sites/review-site-hierarchy.tsx
aryamanshodhan, line 1: Question: Why are we using start date and end date here? Is that because we no longer need a hook to get month keys, since we are now moving to collection cycles?

src/lib/storage-keys.ts
```diff
@@ -5,7 +5,7 @@ export const StorageKeys = {
         endMonth: 'review.endMonth',
         selectedLocation: 'review.selectedLocation',
         expandedSitePaths: 'review.expandedSitePaths',
-        collapsedMonths: 'review.collapsedMonths',
+        collapsedSegments: 'review.collapsedSegments',
```
aryamanshodhan, line 8: What is a collapsedSegment?

src/features/review/sites-list/components/sites/review-sites-list.tsx
```diff
@@ -13,20 +13,28 @@ import {
...
+import ReviewSiteHierarchy from '@/features/review/sites-list/components/sites/review-site-hierarchy';
+import { type ReviewSiteSessionSummary } from '@/features/review/sites-list/utils/review-site-session-summary';
+import { useTranslations } from 'next-intl';
 
 interface ReviewSiteListProps {
     sites: Site[];
     locationQueryParam: LocationQueryParam;
     startMonth: Date;
     endMonth: Date;
+    isCollectionCyclesPending: boolean;
```
aryamanshodhan: I dont think this is needed.

src/features/review/sites-list/components/sites/review-sites-list.tsx
```diff
@@ -52,50 +64,45 @@ export default function ReviewSitesList({
             type: 'SURVEILLANCE',
         });
 
+    const isCycleMode = collectionCycles.length > 0;
     const months = eachMonthOfInterval({ start: startMonth, end: endMonth });
 
     const sessionSummariesByMonth = useMemo(() => {
         const map = new Map<string, Map<number, ReviewSiteSessionSummary>>();
-        if (!getAllSessionsResult?.ok) return map;
+        if (isCycleMode || !getAllSessionsResult?.ok) return map;
```
aryamanshodhan: Does this isCycleMode check belong here? It should be whenever we define the UI conditionally no?

src/features/review/sites-list/components/sites/review-sites-list.tsx
```diff
@@ -52,50 +64,45 @@ export default function ReviewSitesList({
...
+                accumulateSessionSummary(monthMap.get(session.siteId), session),
+            );
         }
 
         return map;
-    }, [getAllSessionsResult]);
+    }, [getAllSessionsResult, isCycleMode]);
+
+    const allCycleSegments = useMemo(() => {
+        if (!isCycleMode || !getAllSessionsResult?.ok) return [];
```
aryamanshodhan: Same here. I dont think the isCycleMode check is needed, because you are already gating at render time.

src/features/review/sites-list/components/layout/collection-cycle-picker.tsx
```diff
@@ -0,0 +1,65 @@
...
+    collectionCycles,
+    selectedCycleIds,
+    onChange,
+    disabled,
+}: CollectionCyclePickerProps) {
+    const t = useTranslations('CollectionCycle');
+
+    const selectedCycleIdStrings = selectedCycleIds.map(String);
+
+    function handleStringIdsChange(stringIds: string[]) {
```
aryamanshodhan: Let's call this handleCollectionCycleIdsChange or handleCycleIdsChange

src/features/review/sites-list/utils/format-collection-cycle-label.ts
```diff
@@ -0,0 +1,11 @@
+import type { CollectionCycle } from '@/api/collection-cycle/validation/collection-cycle-schema';
+import { format } from 'date-fns';
+
+export function formatCollectionCycleLabel(
```
aryamanshodhan, line 4: How many consumers does this function have? If > 1, this makes sense. Also in this case, the "t" argument is difficult to follow. Is there opportunity to lift the t one layer above? At that point, is this even needed?

src/features/review/sites-list/components/review-sites-list-page-client.tsx
```diff
@@ -128,19 +168,35 @@ export default function ReviewSitesListPageClient() {
...
+                                }
+                            />
+                            <MonthRangePicker
+                                startMonth={startMonth}
+                                endMonth={endMonth}
+                                onStartMonthChange={handleStartMonthChange}
+                                onEndMonthChange={handleEndMonthChange}
+                                maxDate={new Date()}
+                            />
+                        </div>
```
aryamanshodhan: I would move these into the header (CollectionCyclePicker + MonthRangePicker)

src/features/review/sites-list/components/review-sites-list-page-client.tsx
```diff
@@ -78,22 +82,58 @@ export default function ReviewSitesListPageClient() {
...
+    const collectionCycles = getCollectionCyclesResult?.ok
+        ? getCollectionCyclesResult.data.collectionCycles
+        : [];
+
+    const isCycleMode = collectionCycles.length > 0;
+
     useEffect(() => {
         setExpandedSitePaths(new Set());
-        setCollapsedMonths(new Set());
+        setCollapsedSegments(new Set());
```
aryamanshodhan, line 107: Semantically, does it make sense to rename this to setExpandedSegments(new Set())?

src/features/review/sites-list/utils/build-collection-cycle-segments.ts
```diff
@@ -0,0 +1,50 @@
+import type { CollectionCycle } from '@/api/collection-cycle/validation/collection-cycle-schema';
+import type { Session } from '@/api/session/validation/session-schema';
+import type { ReviewSiteSessionSummary } from '@/features/review/sites-list/utils/review-site-session-summary';
+import { accumulateSessionSummary } from '@/features/review/sites-list/utils/accumulate-session-summary';
+
+export interface CollectionCycleSegment {
```
aryamanshodhan: I would push back on whether we need an additional intermediate interface. Is it absolutely necessary? Reconsider.

src/features/review/sites-list/utils/build-collection-cycle-segments.ts
```diff
@@ -0,0 +1,50 @@
...
+    cycle: CollectionCycle | null;
+    sessionSummaryBySiteId: Map<number, ReviewSiteSessionSummary>;
+}
+
+export function buildCollectionCycleSegments(
```
aryamanshodhan, line 6: If there is only one consumer of this, should we inline it?

aryamanshodhan (changes requested):
Minor change recommendations/questions

src/features/review/sites-list/components/sites/review-site-hierarchy.tsx
csmanel: Correct, since cycles don't map to calendar months, we standardized on startDate/endDate for both modes so the component works the same way regardless.

src/lib/storage-keys.ts
csmanel: A "segment" is a collapsible time-based section in the Review list, either a Collection Cycle or a calendar month (for programs without a schedule). Same term used in buildCollectionCycleSegments.

src/features/review/sites-list/utils/format-collection-cycle-label.ts
csmanel: 2 consumers: CollectionCyclePicker and ReviewSitesList. 
The t arg is necessary, I don't beleive you can call useTranslations in a utility function, so both callers pass it in.

src/features/review/sites-list/components/review-sites-list-page-client.tsx
csmanel: I don't think so, these track different things. CollapsedGroups is which time-based sections are collapsed, expandedSitePaths is which nodes in the site hierarchy tree are expanded within a section. They reset together but are otherwise independent.

src/api/collection-cycle/validation/get-collection-cycles-schema.ts
```diff
@@ -0,0 +1,20 @@
+import { z } from 'zod';
+import { collectionCycleSchema } from '@/api/collection-cycle/validation/collection-cycle-schema';
+
+export const getCollectionCyclesQueryParamsSchema = z.object({
```
aryamanshodhan, line 4: API is structured incorrectly. ProgramId is not a query parameter. Please fix all the API files to match the backend get endpoint:

GET /programs/{program_id}/collection-cycles

aryamanshodhan (changes requested):

aryamanshodhan (approved):

PR #136, opened 2026-06-03
docsite: test run docsite

csmanel:


patches/nextra-theme-docs+4.6.1.patch
aryamanshodhan, line 1: Question: Do we need this file?

src/proxy.ts
```diff
@@ -15,7 +15,11 @@ export async function proxy(request: NextRequest) {
         return NextResponse.redirect(new URL('/', request.url));
     }
 
-    if (!accessToken && !PUBLIC_ROUTES.has(pathname)) {
+    if (
+        !accessToken &&
+        !PUBLIC_ROUTES.has(pathname) &&
```
aryamanshodhan: You can just add /docs to the public routes instead.

mdx-components.tsx
aryamanshodhan, line 1: Does this file need to be at the root? OR can it be moved into src/hooks?

aryamanshodhan (changes requested):
Extremely minor comments

aryamanshodhan (approved):

PR #140, opened 2026-06-09
VCV-198: Specimen composition tab needs to filter the sex and and status graphs by species

scho324:
* One filter for all graphs on the specimen composition tab
* The multi-select dropdown only includes species that were detected at some point within the selected time range as options
* Excludes non-mosquito/non mosquito by default


src/features/operations/specimen-composition/components/operations-specimen-composition.tsx
```diff
@@ -49,27 +65,78 @@ export default function OperationsSpecimenComposition({
...
+        const validOptions = new Set(speciesOptions);
+        const species =
+            storedSpecies ??
+            speciesOptions.filter(species => !isNonMosquito(species));
+        return species.filter(s => validOptions.has(s));
+    }, [storedSpecies, speciesOptions]);
+
     if (isGetMonthlySpecimensCountPending || !getMonthlySpecimensCountResult) {
-        return <h1>LOADING...</h1>;
+        return <h1>{t('capsLoading')}</h1>;
```
aryamanshodhan: capsLoading is not an intuitive key

src/features/operations/specimen-composition/components/operations-specimen-composition.tsx
```diff
@@ -49,27 +65,78 @@ export default function OperationsSpecimenComposition({
...
     }
 
     if (!getMonthlySpecimensCountResult.ok) {
-        return <h1>ERROR: {getMonthlySpecimensCountResult.error.message}</h1>;
+        return (
+            <h1>
+                {t('capsError')}
```
aryamanshodhan: capsError is not an intuitive key

src/features/operations/specimen-composition/utils/specimen-composition-helpers.ts
```diff
@@ -41,44 +47,109 @@ const CHART_COLOR_FAMILY_BY_CLASSIFICATION_AXIS: Record<
...
+        return a.localeCompare(b);
+    });
+}
+
+function aggregateCountsBySpecies(
+    specimenClassificationAxis: SpecimenClassificationAxis,
+    monthlySpecimenCount: GetMonthlySpecimensCountSuccessPayload['data'][number],
+    selectedSpecies: string[],
+): Record<string, number> {
+    const bySpecies: Record<
```
aryamanshodhan: This variable name is not intuitive. What is this map storing?

src/features/operations/specimen-composition/utils/specimen-composition-helpers.ts
```diff
@@ -41,44 +47,109 @@ const CHART_COLOR_FAMILY_BY_CLASSIFICATION_AXIS: Record<
...
+        string,
+        Record<string, number>
+    > = specimenClassificationAxis === 'species'
+        ? Object.fromEntries(
+              Object.entries(monthlySpecimenCount.species).map(
+                  ([species, count]) => [species, { [species]: count }],
+              ),
+          )
+        : specimenClassificationAxis === 'sex'
```
aryamanshodhan: Question: Why are the classification axes only for species and sex?

src/features/operations/specimen-composition/components/operations-specimen-composition.tsx
```diff
@@ -49,27 +65,78 @@ export default function OperationsSpecimenComposition({
...
+    const speciesOptions = useMemo(() => {
+        if (!getMonthlySpecimensCountResult?.ok) return [];
+        return getSpeciesOptions(getMonthlySpecimensCountResult.data.data);
+    }, [getMonthlySpecimensCountResult]);
+
+    const validSelectedSpecies = useMemo(() => {
+        const validOptions = new Set(speciesOptions);
+        const species =
+            storedSpecies ??
+            speciesOptions.filter(species => !isNonMosquito(species));
```
aryamanshodhan, line 81: Explain this condition. Why are we checking for only mosquitoes?

aryamanshodhan (changes requested):
Looks good Sophia, but there are a couple of things that need to be addressed

src/features/operations/specimen-composition/utils/specimen-composition-helpers.ts
scho324: Sex and abdomen status will need by-species counts; this is a check for whether the axis is species so it can be reformatted to match the sex/abdomen status by-species counts structure for simpler/more consistent handling. Otherwise, it checks for whether the axis is sex; if it's neither, then the axis is the abdomen status.

src/features/operations/specimen-composition/components/operations-specimen-composition.tsx
scho324: If the user hasn't yet selected a set of species from the filter, the default/initial selection is to have everything but non-mosquitos.

src/features/operations/specimen-composition/utils/specimen-composition-helpers.ts
scho324: Counts formatted as
{species A: { sex A: 123, sex B: 123 }, species B: (etc.)} 

Will adjust the code so that reformatting is no longer needed

aryamanshodhan (dismissed):
Optional modification: I would recommend removing the "allCaps" part from the translation key. Instead, you could indicate which file it is coming from, or what loading and error state it is for.

aryamanshodhan (approved):

PR #147, opened 2026-06-18
VCV-233: update whitelist page and forbidden page

csmanel:


aryamanshodhan (approved):
Looks good to me! Just keep in mind, right now if I remove the reason query param, I am able to revisit the forbidden page by directly modifying the URL. Does not need an immediate fix, but we should think about it later.

PR #148, opened 2026-06-23
VCV-211: Email verification

scho324:
* Upon signup (or login for existing accounts), if user is not verified, prompts user to send a verification email
* Verification email is sent to the email that created the account 
* Upon email verification, user is redirected to dashboard
* Already-verified emails are redirected to dashboard upon login 
* Redirects to /login with the token in the URL if the user clicks on it without being logged in, then retries verifying the email
* The resend verification email button on expired/invalid token redirects to /login to show the verification email prompt 
* Email-verified cookie is set if the verification check on login has emailVerified = true or when the verification is successful


src/proxy.ts
```diff
@@ -1,29 +1,46 @@
...
+    if (!accessToken && isAuthOnly) {
+        const loginUrl = new URL('/login', request.url);
+        loginUrl.searchParams.set(
+            'redirect',
+            request.nextUrl.pathname + request.nextUrl.search,
+        );
+        return NextResponse.redirect(loginUrl);
+    }
+
     return NextResponse.next();
```
csmanel, line 38: What is email verification actually doing. In the scenario where a user is whitelisted, but not verified should they still have access to protected routes? 

At the moment, if I create an account, whitelist it (I tested with permission level 4), I can navigate to any routes I want. Should we be blocking protected routes from non verified users?

src/features/auth/components/login-form.tsx
```diff
@@ -20,10 +20,15 @@ import { Button } from '@/components/ui/button';
...
 import { Eye, Lock, Mail } from 'lucide-react';
 import { useState } from 'react';
+import type { GetUserProfileSuccessPayload } from '@/api/user/validation/get-user-profile-schema';
+import { useSearchParams } from 'next/navigation';
+import { useTranslations } from 'next-intl';
 
 export default function LoginForm() {
     const router = useRouter();
     const [showPassword, setShowPassword] = useState(false);
+    const callbackUrl = useSearchParams().get('callbackUrl');
```
csmanel: Potential bug: if the user is logged out when they click the email link, they never get sent back to finish verifying.

The proxy saves the destination under 
`redirect`:`/login?redirect=/verify-email?token=abc123`

But this line reads `callbackUrl`:

```ts
const callbackUrl = useSearchParams().get('callbackUrl'); //potentially empty

File: src/features/auth/components/email-verification-prompt.tsx [hunk 3]
```diff
@@ -0,0 +1,76 @@
...
+                case 400:
+                    router.replace('/');
+                    router.refresh();
+                    return;
+                case 401:
+                    router.replace('/login');
+                    router.refresh();
+                    return;
+                default:
+                    return (
```
csmanel: What happens on a 500/network error?
 
If the user doesn't see error UI as expected potentially we should set error state here to render some message/retry button?

src/app/(auth)/verify-email/page.tsx
```diff
@@ -0,0 +1,90 @@
...
+                                {t('resendVerificationEmailButton')}
+                            </Button>
+                        </>
+                    ) : response.error.message ===
+                      'Invalid or expired verification token' ? (
+                        <>
+                            <p className="text-muted-foreground text-center text-sm">
+                                {t('invalidOrExpiredVerificationLink')}
+                            </p>
+                            <Button onClick={resendVerificationCode}>
```
csmanel: potential issue, a user arriving from an expired/invalid verification link sees a "Resend verification email" button that looks functional but is a silent dead end. 

I believe we have an existing pattern for this, [src/app/forbidden/page.tsx](vscode-webview://1q3o91sqkpp8rvrn3u3jugt22h12lqjtbqhidnanfrke3mp3qk57/src/app/forbidden/page.tsx) (also an async Server Component), which navigates via <Button asChild><Link> 

```
<Button asChild>
    <Link href="/login">{t('resendVerificationEmailButton')}</Link>
</Button>
```

Then we also likely don't need the `resendVerificationCode`

src/app/(auth)/signup/page.tsx
```diff
@@ -1,26 +1,46 @@
...
+                        {t('loginLinkText')}
+                    </Link>
+                </p>
+            </AuthShell>
+        );
+    } else {
+        return (
+            <AuthShell
+                title={t('verifyYourEmail')}
+                description={'emailVerificationPromptDescription'}
```
csmanel: very minor, this should be: `description={t('emailVerificationPromptDescription')}`

csmanel (changes requested):
Overall looking fairly solid! 

Just a few minor things to consider/look into!

src/api/auth/validation/send-verification-email-schema.ts
aryamanshodhan, line 1: rename this file according to how other endpoints have been named. postSendEmailVerification

post-send-email-verification-schema.ts

src/api/auth/validation/send-verification-email-schema.ts
```diff
@@ -0,0 +1,11 @@
+import { z } from 'zod';
+
+export const verificationEmailResponseSchema = z.object({
+    message: z.string(),
+});
+
+export type VerificationEmailResponseBody = z.infer<
```
aryamanshodhan: PostSendEmailVerificationResponseBody

src/api/auth/validation/send-verification-email-schema.ts
```diff
@@ -0,0 +1,11 @@
...
+    typeof verificationEmailResponseSchema
+>;
+
+export type VerificationEmailSuccessPayload = VerificationEmailResponseBody;
```
aryamanshodhan: PostSendEmailVerificationSuccessPayload

src/api/auth/validation/verify-email-schema.ts
aryamanshodhan, line 1: Mention the type of API request in the name.

post-verify-email-schema.ts

src/api/auth/validation/verify-email-schema.ts
```diff
@@ -0,0 +1,18 @@
+import { z } from 'zod';
+
+export const verifyEmailRequestSchema = z.object({
```
aryamanshodhan: postVerifyEmailRequestSchema

src/api/auth/validation/verify-email-schema.ts
```diff
@@ -0,0 +1,18 @@
...
+    token: z.string(),
+});
+
+export const verifyEmailResponseSchema = z.object({
```
aryamanshodhan: postVerifyEmailResponseSchema

src/api/auth/validation/verify-email-schema.ts
```diff
@@ -0,0 +1,18 @@
...
+    user: z.object({
+        id: z.number(),
+        email: z.email(),
+        emailVerified: z.boolean(),
+    }),
+});
+
+export type VerifyEmailRequestBody = z.infer<typeof verifyEmailRequestSchema>;
+export type VerifyEmailResponseBody = z.infer<typeof verifyEmailResponseSchema>;
+export type VerifyEmailSuccessPayload = VerifyEmailResponseBody;
```
aryamanshodhan: PostVerifyEmailRequestBody
PostVerifyEmailResponseBody
PostVerifyEmailSuccessPayload

src/api/auth/send-verification-email.ts
```diff
@@ -0,0 +1,22 @@
+import { type Result } from '@/lib/result/result';
+import {
+    verificationEmailResponseSchema,
+    type VerificationEmailResponseBody,
+} from '@/api/auth/validation/send-verification-email-schema';
+import type { NetworkError } from '@/lib/network/network-error';
+import { safeApiCall } from '@/lib/network/safe-api-call';
+
+export async function sendVerificationEmail(
```
aryamanshodhan: postSendVerificationEmail

src/api/auth/post-send-verification-email.ts
aryamanshodhan, line 1: Rename file to post-send-verification-email.ts

src/api/auth/verify-email.ts
```diff
@@ -0,0 +1,31 @@
...
+import {
+    verifyEmailRequestSchema,
+    verifyEmailResponseSchema,
+    type VerifyEmailRequestBody,
+    type VerifyEmailResponseBody,
+} from '@/api/auth/validation/verify-email-schema';
+import { err, type Result } from '@/lib/result/result';
+import { safeApiCall } from '@/lib/network/safe-api-call';
+
+export async function verifyEmail(
```
aryamanshodhan: postVerifyEmail

src/api/user/post-verify.ts
aryamanshodhan, line 1: Rename file to post-verify-email.ts

src/app/(auth)/login/page.tsx
```diff
@@ -1,26 +1,46 @@
...
-                >
-                    Create one
-                </Link>
-            </p>
-        </AuthShell>
-    );
+export default async function LoginPage() {
+    const cookieStore = await cookies();
+    const accessToken = cookieStore.get(ACCESS_COOKIE_NAME)?.value;
+    const t = await getTranslations('Auth');
```
aryamanshodhan, line 8: Calling await cookies() from a component is not best practice. The proxy.ts file will ensure that the user is redirected to the appropriate page based on whether they are authenticated or not. You should consider implementing redirect logic over there instead of conditionally rendering on this page based on whether the user has an access token or not, aka whether they are logged in or not. Maybe you could have a separate page for email verification.

src/app/(auth)/signup/page.tsx
```diff
@@ -1,26 +1,46 @@
...
-                    href="/login"
-                    className="text-primary font-medium hover:underline"
-                >
-                    Login instead
-                </Link>
-            </p>
-        </AuthShell>
-    );
+export default async function SignupPage() {
+    const cookieStore = await cookies();
```
aryamanshodhan: same here. proxy.ts already handles redirecting to the appropriate pages. You should avoid conditional rendering here. Make email verification a new page if you want.

src/app/(auth)/verify-email/page.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+export default async function VerifyEmailPage({
+    searchParams,
+}: VerifyEmailPageProps) {
+    const t = await getTranslations('Auth');
+    async function resendVerificationCode() {
+        redirect('/login');
+    }
+
+    const accessToken = (await cookies()).get(ACCESS_COOKIE_NAME)?.value;
```
aryamanshodhan: Use proxy.ts and withAuthSession if needed to know whether the user is authenticated.

src/app/(auth)/verify-email/page.tsx
```diff
@@ -0,0 +1,90 @@
...
+                <p>{t('missingVerificationToken')}</p>
+            </AuthShell>
+        );
+    } else {
+        const response = await verifyEmail(accessToken, {
+            token,
+        });
+
+        if (!response.ok) {
+            console.error(
```
aryamanshodhan: Remove this log message

src/app/(auth)/verify-email/page.tsx
```diff
@@ -0,0 +1,90 @@
...
+            return (
+                <AuthShell
+                    title="Verify your email"
+                    description="Email verification in progress."
+                    imageSrc="/assets/auth/images/Login.png"
+                >
+                    <Separator className="my-6" />
+                    {response.error.message ===
+                    'Verification token is required' ? (
+                        <>
```
aryamanshodhan: <Fragment>

src/app/(auth)/verify-email/page.tsx
```diff
@@ -0,0 +1,90 @@
...
+                            <p className="text-muted-foreground text-center text-sm">
+                                {t('invalidVerificationLink')}
+                            </p>
+                            <Button onClick={resendVerificationCode}>
+                                {t('resendVerificationEmailButton')}
+                            </Button>
+                        </>
+                    ) : response.error.message ===
+                      'Invalid or expired verification token' ? (
+                        <>
```
aryamanshodhan: <Fragment>

src/app/(auth)/verify-email/page.tsx
aryamanshodhan: Good point^^

aryamanshodhan: Name variables consistently. Resend verification code or resend verification email?

src/app/(auth)/verify-email/page.tsx
```diff
@@ -0,0 +1,90 @@
...
+                                {t('invalidOrExpiredVerificationLink')}
+                            </p>
+                            <Button onClick={resendVerificationCode}>
+                                {t('resendVerificationEmailButton')}
+                            </Button>
+                        </>
+                    ) : response.error.message ===
+                      'Verification token does not match the authenticated user' ? (
+                        <p className="text-muted-foreground text-center text-sm">
+                            {t('accountVerificationTokenMismatch')}
```
aryamanshodhan: Consider making all the error messages user friendly

src/app/api/auth/email-verification/route.ts
```diff
@@ -0,0 +1,17 @@
+import { sendVerificationEmail } from '@/api/auth/send-verification-email';
+import { type VerificationEmailResponseBody } from '@/api/auth/validation/send-verification-email-schema';
+import { withAuthSession } from '@/lib/auth-session/with-auth-session';
+import { NextResponse } from 'next/server';
+
+export async function POST() {
```
aryamanshodhan: This comment might not necessarily warrant a change. But just verify that the structure you are using (eg. request, response structure validation followed by returning a result) is addressed.

src/app/api/auth/verification-email/route.ts
```diff
@@ -0,0 +1,17 @@
+import { sendVerificationEmail } from '@/api/auth/send-verification-email';
+import { type VerificationEmailResponseBody } from '@/api/auth/validation/send-verification-email-schema';
+import { withAuthSession } from '@/lib/auth-session/with-auth-session';
+import { NextResponse } from 'next/server';
+
+export async function POST() {
+    const sendVerificationEmailResult =
```
aryamanshodhan: postSendEmailVerificationResult

src/app/api/auth/verify/route.ts
```diff
@@ -0,0 +1,42 @@
...
+    } catch {
+        const requestBodyErrorResult = err({
+            kind: 'client',
+            status: 400,
+            message: 'Invalid JSON body',
+        });
+        return NextResponse.json(requestBodyErrorResult, { status: 400 });
+    }
+
+    const verifyEmailResult: Result<VerifyEmailResponseBody, NetworkError> =
```
aryamanshodhan: postVerifyEmailResult

src/lib/auth-session/cookies.ts
```diff
@@ -31,7 +32,18 @@ export function setRefreshCookie(response: NextResponse, refreshToken: string) {
     });
 }
 
+export function setVerifiedEmailCookie(response: NextResponse) {
```
aryamanshodhan: We do not need to save email verified cookie inside HttpOnly cookies. Email verification is a one time operation, after which the user no longer needs it, because the user profile returns a boolean for whether the user's email is verified. Therefore, you do not need to save this every time.

src/app/api/users/profile/route.ts
```diff
@@ -2,14 +2,24 @@ import { getUserProfile } from '@/api/user/get-user-profile';
...
         await withAuthSession<GetUserProfileResponseBody>(getUserProfile);
 
-    return NextResponse.json(authorizedGetUserProfileResult, {
+    const response = NextResponse.json(authorizedGetUserProfileResult, {
         status: authorizedGetUserProfileResult.ok
             ? 200
             : (authorizedGetUserProfileResult.error.status ?? 400),
     });
+
+    if (
```
aryamanshodhan: You do not need this block I think. check out the comment within cookies.ts

src/lib/network/safe-api-call.ts
```diff
@@ -22,10 +22,14 @@ export async function safeApiCall<T>(
     options?: RequestInit,
     validationSchema?: z.ZodType<T>,
 ): Promise<Result<T, NetworkError>> {
+    const method = options?.method?.toUpperCase() ?? 'GET';
+    const needsBody = ['POST', 'PUT', 'PATCH'].includes(method);
+    const body = needsBody ? (options?.body ?? '{}') : options?.body;
```
aryamanshodhan: Explain the need for this. You should not need this, because we already have endpoints which pass in a body without the need for a separate body. options?: RequestInit already handles the passing in of a body. You should make sure to check out examples of existing similar use cases/implementations within the codebase.

src/proxy.ts
```diff
@@ -1,29 +1,46 @@
...
 export async function proxy(request: NextRequest) {
     const pathname = request.nextUrl.pathname;
     const accessToken = request.cookies.get(ACCESS_COOKIE_NAME)?.value;
+    const emailVerified =
+        request.cookies.get(VERIFIED_COOKIE_NAME)?.value === 'true';
+
+    const isPublic =
+        PUBLIC_ROUTES.has(pathname) ||
+        PUBLIC_PREFIXES.some(prefix => pathname.startsWith(prefix));
+    const isAuthOnly = AUTH_ONLY_ROUTES.has(pathname);
```
aryamanshodhan, line 15: What does isAuthOnly do?

src/proxy.ts
aryamanshodhan: Yes, we should

src/features/auth/components/login-form.tsx
aryamanshodhan: What is a callbackUrl?

src/features/auth/components/email-verification-page-client.tsx
```diff
@@ -0,0 +1,76 @@
...
+import { useRouter } from 'next/navigation';
+import { useState } from 'react';
+import { useTranslations } from 'next-intl';
+
+export default function EmailVerificationPrompt() {
+    const [submitted, setSubmitted] = useState(false);
+    const router = useRouter();
+    const t = useTranslations('Auth');
+
+    async function sendEmail() {
```
aryamanshodhan, line 20: This should be within a hook function, and you should use tanstack query for this if you are calling from within a component.

src/features/auth/components/email-verification-prompt.tsx
```diff
@@ -0,0 +1,76 @@
...
+    const router = useRouter();
+    const t = useTranslations('Auth');
+
+    async function sendEmail() {
+        const response = await fetch('/api/auth/verification-email', {
+            method: 'POST',
+            credentials: 'include',
+        });
+
+        const sendEmailResult: Result<
```
aryamanshodhan: postSendEmailVerificationResult

src/features/auth/components/email-verification-prompt.tsx
```diff
@@ -0,0 +1,76 @@
...
+            VerificationEmailSuccessPayload,
+            NetworkError
+        > = await response.json();
+
+        if (!response.ok || !sendEmailResult.ok) {
+            console.error(
```
aryamanshodhan: Remove this error log

src/features/auth/components/email-verification-prompt.tsx
```diff
@@ -0,0 +1,76 @@
...
+                'Sending Verification Email Failed: ',
+                response.statusText,
+            );
+            switch (response.status) {
+                case 400:
+                    router.replace('/');
```
aryamanshodhan: This should be handled within a page client.

src/features/auth/components/email-verification-prompt.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+    if (!submitted) {
+        return (
+            <Button className="w-full" onClick={sendEmail}>
+                {t('verifyEmailButton')}
+            </Button>
+        );
+    } else {
+        return (
+            <>
```
aryamanshodhan: `<Fragment>`

aryamanshodhan (changes requested):
Great start Sophia! However, there are some important changes needed. Please address the comments and submit an updated PR.

src/lib/network/safe-api-call.ts
scho324: safeApiCall automatically puts in 'Content-Type': 'application/json', and throws the error response "Body cannot be empty when content-type is set to 'application/json'", but POST users/email/send-verification doesn't take a body. I just now set it to only include the content-type if options?: RequestInit includes a body.

aryamanshodhan: Can you not pass in an empty request body from the call site?

src/lib/auth-session/cookies.ts
scho324: What if the user logs in and isn't verified, then tries to go to the dashboard by removing /login from the URL? Wouldn't the proxy have to check their verification/fetch the user profile again anyway to know whether to redirect them to the verification email prompt?

src/lib/network/safe-api-call.ts
scho324: No, it still gave that error response when I tried it.

src/app/(auth)/verify-email/page.tsx
scho324: Are the current error messages in en.json not user friendly enough? A lot of the 'common errors' seem to have the same status code based on the file Cole sent from Bowen, so this was the only way I could think of to distinguish them. 

 Status | Error | Frontend action |
|--------|-------|-----------------|
| 400 | `Verification token is required` | Show invalid link message |
| 400 | `Invalid or expired verification token` | Prompt to resend verification email |
| 400 | `Verification token does not match the authenticated user` | Ask user to log in as the account that received the email |
| 401 | `Unauthorized: User authentication required` | Redirect to login, then retry with same `token` in URL |

src/proxy.ts
scho324: Currently checks if the page only requires authorization but not email verification - either the email verification prompt page, or the verify email page

csmanel (dismissed):
Seemed to work for me, do a final round of testing please!

csmanel (commented):
I am curious on your thoughts, I see two potential issues at the moment. 
I can navigate to /verify-email, /email-verification by url at the moment, and then I can interact with the verify email button (which if I am already verified I am not sure that I should be able to).

Could we do something like this, In `email-verification/page.tsx` (and `verify-email/page.tsx`), make a `getUserProfile` call server-side before rendering (following the pattern for `isWhitelisted` in (dashboard)/layout.tsx) and then we can just redirect if `emailVerified` is true 

I think this would potentially mean we can reduce our cookie logic? 

Or, potentially in `(dashboard)/layout.tsx`, right after the existing `getUserProfile` call (near the isWhitelisted check), add:

```
if (userProfile.emailVerified) {
    setVerifiedEmailCookie(response);
}
```

Let me know what you think, regardless we need some form of check for if a user is already verified that way they cannot interact with the email verification route.

src/app/(auth)/verify-email/page.tsx
```diff
@@ -0,0 +1,103 @@
...
+                            <Fragment>
+                                <p className="text-muted-foreground text-center text-sm">
+                                    {t('accountVerificationTokenMismatch')}
+                                </p>
+                                <LogoutButton />
+                            </Fragment>
+                        ) : (
+                            <Fragment>
+                                <p className="text-muted-foreground text-center text-sm">
+                                    response.error.message
```
csmanel: should this have { } ?

src/app/(auth)/verify-email/page.tsx
```diff
@@ -0,0 +1,103 @@
...
+        if (!response.ok) {
+            return (
+                <AuthShell
+                    title={t('verifyYourEmail')}
+                    description={t('verifyEmailIssueDescription')}
+                    imageSrc="/assets/auth/images/Login.png"
+                >
+                    <Separator className="my-6" />
+                    <div className="flex flex-col items-center gap-4">
+                        {response.error.message ===
```
csmanel, line 73: Is this potentially fragile? Matching on response.error.message strings directly. 
I believe other places in the codebase discriminate errors via kind/status instead (checkout `metadata-review-workspace` or `(dashboard)/layout`)
Worth double checking if that'd work better here.

scho324: The issue is that it seems like three of the possible errors Bowen described in the specifications have the same status code, I'm not sure what else to handle the 400s with.

| Status | Error | Frontend action |
|--------|-------|-----------------|
| 400 | `Verification token is required` | Show invalid link message |
| 400 | `Invalid or expired verification token` | Prompt to resend verification email |
| 400 | `Verification token does not match the authenticated user` | Ask user to log in as the account that received the email |
| 401 | `Unauthorized: User authentication required` | Redirect to login, then retry with same `token` in URL |

csmanel: I see, that definitely changes things lol

We may just have to use message then.
@aryamanshodhan, I would be curious on your thoughts. 
Maybe we continue this way, but ask Bowen for distinct codes so we can change it in the future?

src/app/(auth)/email-verification/page.tsx
```diff
@@ -0,0 +1,17 @@
...
+export default async function EmailVerificationPage() {
+    const t = await getTranslations('Auth');
+
+    return (
+        <AuthShell
+            title={t('verifyYourEmail')}
+            description={t('emailVerificationPromptDescription')}
+            imageSrc="/assets/auth/images/Login.png"
+        >
+            <EmailVerificationPrompt />
```
aryamanshodhan: This must be within a page client. Have you defined a page client for this new page? Make sure the component hierarchy is consistent throughout the codebase.

src/app/(auth)/verify-email/page.tsx
aryamanshodhan: I'd say display the errors based on the workflow and the stage the user is in. The 400 error is triggered as a result of a "bad request" by the user. Even though the error messages are different, investigate the potential places from where each error case might be thrown, and just display based on the status code, and trust that the backend will return the appropriate error message. Eg. 400 and 401 errors happen for different reason. Just display based on the error code, and trust that the backend will return the correct 400 error message based on the situation.

src/app/api/auth/email-verification/route.ts
aryamanshodhan, line 1: Keep the path consistent with the upstream. There you had /users/email/send-verification

Might need to move this under the users/ route system

src/app/api/users/verify/route.ts
aryamanshodhan, line 1: Same here. Make sure the route system matches the backend.

src/components/auth-session/logout-button.tsx
```diff
@@ -4,12 +4,15 @@ import { Button } from '@/components/ui/button';
...
 import { LogOut } from 'lucide-react';
+import { clearResendCooldown } from '@/lib/hooks/use-resend-cooldown';
+import { StorageKeys } from '@/lib/storage-keys';
 
 export default function LogoutButton() {
     const queryClient = useQueryClient();
 
     async function handleLogout() {
         queryClient.clear();
+        clearResendCooldown(StorageKeys.auth.verificationEmailSentTimestamp);
```
aryamanshodhan: Are we changing local storage to nuqs? @csmanel need your input here.

src/lib/network/safe-api-call.ts
```diff
@@ -22,13 +22,14 @@ export async function safeApiCall<T>(
...
 ): Promise<Result<T, NetworkError>> {
+    const hasBody = options?.body !== undefined;
     let response: Response;
     try {
         response = await fetch(constructUrl(path), {
             ...options,
             cache: 'no-cache',
             headers: {
-                'Content-Type': 'application/json',
+                ...(hasBody ? { 'Content-Type': 'application/json' } : {}),
```
aryamanshodhan, line 32: Explain why you need this again?

aryamanshodhan (changes requested):
Requires a few minor changes. Please also convince yourself that the user is not getting soft locked for any reason.

src/lib/network/safe-api-call.ts
scho324: The send-verification endpoint doesn't take a body, and making the request gives an error if 'Content-Type': 'application/json' gets added as a header because it expects a body that doesn't exist.

src/app/(auth)/verify-email/page.tsx
csmanel: Would something like this work? 
We only care about if we need to block the user from re-sending another verification right? So we can look for specifically that error message and anything else we can trust that the backend will return something appropriate.

```js
if (!response.ok) {
    return (
        <AuthShell
            title={t('verifyYourEmail')}
            description={t('verifyEmailIssueDescription')}
            imageSrc="/assets/auth/images/Login.png"
        >
            <Separator className="my-6" />
            <div className="flex flex-col items-center gap-4">
                {response.error.message === VERIFICATION_TOKEN_MISMATCH_ERROR ? (
                    <Fragment>
                        <p className="text-muted-foreground text-center text-sm">
                            {t('accountVerificationTokenMismatch')}
                        </p>
                        <LogoutButton />
                    </Fragment>
                ) : (
                    <Fragment>
                        <p className="text-muted-foreground text-center text-sm">
                            {response.error.message ?? t('somethingWentWrong')}
                        </p>
                        <Button className="w-full">
                            <Link href="/email-verification">
                                {t('resendVerificationEmailButton')}
                            </Link>
                        </Button>
                        <LogoutButton />
                    </Fragment>
                )}
            </div>
        </AuthShell>
    );
} else {
    redirect('/');
}
```

src/app/(auth)/verify-email/page.tsx
```diff
@@ -0,0 +1,104 @@
...
+import { getTranslations } from 'next-intl/server';
+import { Separator } from '@/components/ui/separator';
+import Link from 'next/link';
+import { withAuthSession } from '@/lib/auth-session/with-auth-session';
+import { Fragment } from 'react/jsx-runtime';
+import LogoutButton from '@/components/auth-session/logout-button';
+import ReturnHomeButton from '@/components/auth-session/return-home-button';
+
+const TOKEN_MISMATCH_ERROR_MESSAGE =
+    'Verification token does not match the authenticated user';
```
aryamanshodhan, line 15: messages.json?

src/components/auth-session/logout-button.tsx
```diff
@@ -4,12 +4,14 @@ import { Button } from '@/components/ui/button';
...
 import { logout } from '@/lib/auth-session/logout';
 import { LogOut } from 'lucide-react';
+import { clearResendCooldown } from '@/lib/hooks/use-resend-cooldown';
 
 export default function LogoutButton() {
     const queryClient = useQueryClient();
 
     async function handleLogout() {
         queryClient.clear();
+        clearResendCooldown();
```
aryamanshodhan, line 14: what are you resending? This naming is not clear

src/components/auth-session/return-home-button.tsx
```diff
@@ -0,0 +1,10 @@
+import { Button } from '@/components/ui/button';
+import Link from 'next/link';
+
+export default function ReturnHomeButton({ label }: { label: string }) {
+    return (
+        <Button asChild variant="outline" className="w-full">
+            <Link href="/">{label}</Link>
+        </Button>
+    );
+}
```
aryamanshodhan, line 10: This component is small enough that it does not need to be its own file, unless there are more than 2 call sites.

src/features/auth/components/login-form.tsx
```diff
@@ -51,6 +57,31 @@ export default function LoginForm() {
             return;
         }
 
+        const userProfileResponse = await fetch('api/users/profile', {
```
aryamanshodhan, line 60: This should be using the hook version of the GET profile endpoint no?

src/lib/hooks/use-resend-cooldown.ts
```diff
@@ -0,0 +1,53 @@
...
+        const rawTime = window.localStorage.getItem(
+            VERIFICATION_EMAIL_COOLDOWN_KEY,
+        );
+        if (!rawTime) return null;
+        const parsed = Number(rawTime);
+        return Number.isFinite(parsed) ? parsed : null;
+    }, []);
+
+    useEffect(() => {
+        function tick() {
```
aryamanshodhan, line 20: We should consider making this a common utility at some point. This is also done during the raw export download link expiration time.

src/proxy.ts
aryamanshodhan: I think we might need to reorder some of these conditions at some point to make things more intuitive.

aryamanshodhan (dismissed):

csmanel (dismissed):

csmanel (approved):

PR #154, opened 2026-06-25
VCV-237: collection timezone display

csmanel:
### Summary
The metadata review table renders each Session's collection date in the **timezone of the region where collection happened** (denormalized onto the Collection Cycle), never the reviewer's browser timezone. Previously the metadata workspace obtained that timezone by re-fetching **every** Collection Cycle for the program to build a `cycleId → timezone` map. Since a review workspace is scoped to one Site + one Cycle (or one month), exactly one timezone is ever in play — so this swaps the redundant fetch for a single `timezone` value carried on the review-detail URL, the same channel that already carries `startDate`, `endDate`, `collectionCycleId`, and `displayName`.

### Changes
- **Link builder** (`review-site-leaf-rows`): appends a `timezone` query param when a non-null value exists; threaded down through `review-site-hierarchy` and `review-sites-list`. Cycle mode passes `cycle.timezone`; month mode passes nothing (param omitted → UTC).
- **Detail page** (`page.tsx` → `review-site-details-page-client`): reads the optional `timezone` searchParam and forwards it as a prop to all three review workspaces (metadata, image, certification).
- **Metadata workspace**: removed the `useGetCollectionCycles` query, the `useGetUserPermissions`/`programId` derivation, and the `cycleId → timezone` map; now passes a single `timezone` down.
- **Metadata table**: replaced the `timezoneByCycleId` map prop with a single `timezone: string | null`, used directly in `formatDateInTimezone`.
- **Robustness guard** (`getValidTimezone` in `format-date-in-timezone`): coerces a malformed (non-IANA) timezone to UTC instead of throwing. Protects every caller (sites list + detail page), not just the new URL path.

### Behavior / fallback
- Valid IANA zone → dates render in that zone.
- `timezone` omitted, empty, or `null` → UTC display (matches documented null-timezone behavior; a stripped/shared link is safe, never an error).
- Malformed zone string → UTC display (previously crashed the page — see note below).
- Cycle membership is unaffected: it stays backend-owned (instant comparison), timezone-independent. Timezone is display-only.

### Testing
No automated tests added — this is prop/URL plumbing on top of the already-exercised `formatDateInTimezone` helper, typed end to end. Verified manually by driving the running app:
- Cycle-mode site → review link carries `timezone=Africa/Kampala`; detail URL preserves it.
- Metadata table dates render in the cycle's zone — confirmed by a 25-hour timezone swing flipping the displayed day (May 27 ↔ May 26), proving the value is actually consumed.
- **Zero `collection-cycles` requests** fire on the detail page across every load (the removed fetch).
- Omitted / empty / malformed `timezone` params all fall back to UTC with no error.

### Note: backend follow-up (out of scope)
The root cause of the malformed-zone crash is that the backend stores the timezone string **unvalidated** — `POST /collection-schedules` accepts any string (verified in `vectorcam-api`, no IANA check anywhere). The frontend guard added here is defense-in-depth for the user-editable URL vector this PR introduces; the durable fix is IANA validation on the backend write path, tracked separately.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/features/review/sites-list/components/sites/review-site-leaf-rows.tsx
+++ b/src/features/review/sites-list/components/sites/review-site-leaf-rows.tsx
@@ -23,6 +23,7 @@ interface ReviewSiteLeafRowsProps {
     startDate: string;
     endDate: string;
     collectionCycleId?: number;
+    timezone?: string | null;
 }
 
 function buildReviewHref(
@@ -31,6 +32,7 @@ function buildReviewHref(
     endDate: string,
     displayName: string,
     collectionCycleId?: number,
+    timezone?: string | null,
 ) {
     const queryParams = new URLSearchParams({
         startDate,
@@ -39,6 +41,9 @@ function buildReviewHref(
     });
     if (collectionCycleId !== undefined) {
         queryParams.set('collectionCycleId', String(collectionCycleId));
+    }
+    if (timezone !== undefined && timezone !== null) {
+        queryParams.set('timezone', timezone);
     }
     return `/review/${site.siteId}?${queryParams.toString()}`;
 }
@@ -50,6 +55,7 @@ export default function ReviewSiteLeafRows({
     startDate,
     endDate,
     collectionCycleId,
+    timezone,
 }: ReviewSiteLeafRowsProps) {
     const t = useTranslations('ReviewSitesList');
     if (sites.length === 0) return null;
@@ -133,6 +139,7 @@ export default function ReviewSiteLeafRows({
                                 endDate,
                                 getDisplayName(site),
                                 collectionCycleId,
+                                timezone,
                             )}
                             className={rowClassName}
                         >
--- a/src/features/review/sites-list/components/sites/review-site-leaf-rows.tsx
+++ b/src/features/review/sites-list/components/sites/review-site-leaf-rows.tsx
@@ -42,7 +42,7 @@ function buildReviewHref(
     if (collectionCycleId !== undefined) {
         queryParams.set('collectionCycleId', String(collectionCycleId));
     }
-    if (timezone !== undefined && timezone !== null) {
+    if (timezone != null) {
         queryParams.set('timezone', timezone);
     }
     return `/review/${site.siteId}?${queryParams.toString()}`;
--- a/src/features/review/sites-list/components/sites/review-sites-list.tsx
+++ b/src/features/review/sites-list/components/sites/review-sites-list.tsx
@@ -216,6 +216,7 @@ export default function ReviewSitesList({
                                     startDate={segmentStartDate}
                                     endDate={segmentEndDate}
                                     collectionCycleId={cycleSegment.cycle.id}
+                                    timezone={cycleSegment.cycle.timezone}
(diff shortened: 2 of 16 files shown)


src/features/review/sites-list/components/sites/review-site-leaf-rows.tsx
```diff
@@ -40,6 +42,9 @@ function buildReviewHref(
     if (collectionCycleId !== undefined) {
         queryParams.set('collectionCycleId', String(collectionCycleId));
     }
+    if (timezone !== undefined && timezone !== null) {
```
aryamanshodhan: This can be collapsed to timezone != null

src/features/review/sites-list/components/sites/review-sites-list.tsx
```diff
@@ -73,8 +74,9 @@ export default function ReviewSitesList({
         if (!getAllSessionsResult?.ok) return map;
 
         for (const session of getAllSessionsResult.data.sessions) {
-            const monthKey = format(
-                new Date(session.collectionDate),
+            const monthKey = formatDateInTimezone(
+                session.collectionDate,
+                'UTC',
```
aryamanshodhan, line 79: Question: Why are we formating in UTC time here? It should be in the local timezone no?

src/utils/format-date-in-timezone.ts
aryamanshodhan, line 1: Lib or utils?

aryamanshodhan (commented):
Looks great Cole! Test this out thoroughly once. Also have questions on a potential bug. Check out the comments

src/features/review/sites-list/components/sites/review-sites-list.tsx
csmanel: If its local wouldn't we run into the same issue again? 
A bordering session within a month not appearing in the right place?

src/utils/format-date-in-timezone.ts
csmanel: I think we can keep it in utils for now? It's a generic leaf helper like cn, and there's only one date util at the moment. 
If we grow more date functionality we can move it to a lib/date folder.

src/features/review/sites-list/components/sites/review-sites-list.tsx
aryamanshodhan: On the other hand that would resolve it no?

Eg. If a session was conducted:
Uganda Time: June 1st 2am
UTC Time: May 31st 11pm
EST Time: May 31st 7pm

On the webapp everywhere it should show in local Uganda time, aka June 1st 2am. If we have it for UTC time, it would show in May instead of June no?

csmanel: I see what you are saying, then should we add timezone to the session/site so month mode can render the correct local time?

csmanel: This is needed while not all programs support collection cycles, but will become unnecessary

aryamanshodhan (approved):

PR #155, opened 2026-06-26
VCV-187: District multiselection dropdown

scho324:
* Geographical summary: sites sorted by district/top-level region 
* Specimen composition + AI performance: aggregated counts 
* Field user compliance: users sorted by district, then by role title


src/features/operations/ai-performance/components/operations-ai-performance.tsx
```diff
@@ -4,27 +4,25 @@ import { useGetAnnotationsSummary } from '@/api/annotation/hooks/use-get-annotat
...
+    selectedLocationNames: string[];
     startDate: string;
     endDate: string;
 }
 
 export default function OperationsAiPerformance({
-    locationQueryParam,
-    selectedLocationName,
+    siteIds,
+    selectedLocationNames,
```
aryamanshodhan: How are you keeping track of which siteIds correspond to which location names?

src/features/operations/components/operations-page-client.tsx
```diff
@@ -70,15 +70,16 @@ export default function OperationsPageClient() {
...
-        selectedLocation,
-        setSelectedLocation,
+        selectedLocations,
+        setSelectedLocations,
         locationTypeName,
         locationDropdownOptions,
-        locationQueryParam,
-        descendantsOfSelectedLocation,
-    } = useLocationSelection(
+        selectedSiteIdsParam,
```
aryamanshodhan, line 80: Explain what this is doing. It seems you found a way to get rid of locationQueryParam altogether. Is this even needed?

src/features/operations/components/operations-page-client.tsx
```diff
@@ -70,15 +70,16 @@ export default function OperationsPageClient() {
...
+        descendantsOfSelectedLocations,
+        siteIdToLocationLabel,
```
aryamanshodhan, line 82: Explain siteIdToLocationLabel

src/features/operations/geographical-summary/components/map-navigator.tsx
```diff
@@ -11,18 +11,18 @@ const SELECTED_LOCATION_ZOOM = 10;
...
-    selectedLocation,
+    selectedLocations,
 }: MapNavigatorProps) {
     const map = useMap();
 
     const { data: geocodeResult } = useGetGeocode(
-        { location: selectedLocation },
-        { enabled: !bounds },
+        { location: selectedLocations[0] ?? '' },
+        { enabled: !bounds && selectedLocations.length <= 1 },
```
aryamanshodhan: Why is this enabled for selected locations.length <= 1? Shouldn't we show it for all locations?

src/lib/location/use-location-multiselection.ts
aryamanshodhan, line 1: If we have useLocationMultiSelection now, does that mean we no longer use useLocationSelection anywhere else in the codebase? If that is the case, can we delete that file if that is unused?

src/lib/location/use-location-multiselection.ts
```diff
@@ -0,0 +1,124 @@
...
+export function useLocationMultiSelection(
+    accessibleSites: Site[],
+    storageKey: string,
+) {
+    const [selectedLocations, setSelectedLocations] = useLocalStorage<string[]>(
+        storageKey,
+        [],
+    );
+
+    const usesLegacyStructure =
```
aryamanshodhan, line 27: Slightly fragile. We are only checking the first site to see if it uses the legacy structure

src/lib/location/use-location-multiselection.ts
aryamanshodhan: Also confirming the structure of this file is similar to the existing use-location-selection.ts file

src/features/operations/geographical-summary/components/marker-info-panel.tsx
```diff
@@ -19,72 +20,100 @@ export default function MarkerInfoPanel({
...
-                                        ({ species, count }) => (
-                                            <p key={species}>
-                                                <span className="mr-1">↳</span>
-                                                {t('siteSpeciesCount', {
-                                                    species,
-                                                    count,
-                                                })}
-                                            </p>
-                                        ),
+            {markersGroupedByLocation.map(([district, groupMarkers]) => (
```
aryamanshodhan: I would not call this district, because we need to ensure that this works for non-uganda programs. Is this the top level location type? What about cases when we do not have districts?

Please ensure this works for non-uganda programs (Ghana). Use these login credentials to verify functionality is as expected:
aryaman-ghana@gmail.com
11111111

aryamanshodhan (changes requested):
This is an extremely clean PR Sophia! Very very well done. Not too many changes, just a few minor questions. I see a lot of instances of locationQueryParam were removed throughout the PR. Please do a secondary pass to ensure there is no dead code. If there is, please remove it. I would also urge you to perform a second pass to make sure that aggregation is happening as expected, eg. counts add up correctly in all views, markers appear as expected, AI performance reflects correctly, report shows the correct rows for all districts after download, etc. Additionally, please add this to the spreadsheet.

src/features/operations/ai-performance/components/operations-ai-performance.tsx
scho324: It's not needed for AI performance since the results are aggregated, so the selectedLocationNames is only passed in for the 'Coverage reflects reviewed specimens for [listed locations] where review means annotated or flagged' and the bit in the interpretation that also uses the location names.

src/features/operations/geographical-summary/components/map-navigator.tsx
scho324: I believe the map seemed to zoom in to the location at selectedLocations[0] before zooming back out to display all the locations' markers once the sites were loaded for marker display. I'm not sure if there is some other thing that can resolve the behavior; now that I'm testing it again, it seems to zoom in to the first location selected anyway since locations are only ever added/removed one at a time.

src/features/operations/geographical-summary/components/marker-info-panel.tsx
scho324: Yes, this is the top-level location type; it works for the Ghana program. 

I have just discovered an error in AI performance where if I am logged in to the Uganda program and select some districts, then log out and then in to the Ghana program to select some locations, it still displays the selected Uganda districts in the list of locations, and vice versa, so I will have to fix that. There is seems to be overlap wherever selectedLocations is directly used.

src/features/operations/ai-performance/components/operations-ai-performance.tsx
aryamanshodhan: You dont need to display all of them. Suppose the user selects 20. You shouldn't list all 20. You can just mention that the data being shown is for the selected locations, and then remove the prop from the component if it is being used just for display.

src/features/operations/geographical-summary/components/map-navigator.tsx
aryamanshodhan: I dont think that is a significant issue. It can zoom into the firts location all it wants as long as we see the markers for all the locations.

src/features/operations/geographical-summary/components/marker-info-panel.tsx
aryamanshodhan: I think this might be an issue where the local storage is not cleared upon logout. This is a bug we need to fix. Please clear the local storage when the user logs out. We should not have cross contamination of data across programs.

src/lib/location/use-location-multiselection.ts
scho324: I believe that useLocationSelection is used by the review page, since you said that only the operations page should have the multiselection dropdown. The structures should be similar, though useLocationSelection still uses LocationQueryParams since it only queries one location at a time and isn't restricted by the endpoints that only allow multiple siteIds and not multiple location names.

src/features/operations/components/operations-page-client.tsx
scho324: Maps siteIds to the corresponding location labels for field team performance/field user compliance so I can find the top-level location label to display in each user's row

src/lib/location/use-location-multiselection.ts
aryamanshodhan: Cool. Then we keep it.

src/features/operations/geographical-summary/components/marker-info-panel.tsx
csmanel: I will push up a fix for useLocalStorage to make sure it clears properly

src/features/operations/components/operations-page-client.tsx
scho324: I think locationQueryParam(s) can be removed from the use-location-multiselection now; it was only left in during implementation for features that hadn't been switched to multiple locations and using siteIds instead of location names.

src/lib/location/use-location-multiselection.ts
```diff
@@ -28,7 +25,7 @@ export function useLocationMultiSelection(
     );
 
     const usesLegacyStructure =
-        accessibleSites.length > 0 && isLegacySite(accessibleSites[0]!);
+        accessibleSites.length > 0 && accessibleSites.some(isLegacySite);
```
aryamanshodhan, line 28: This is not going to be a problem, because any program can only contain either the legacy or hierarchical structure, but in case we encounter some program with both structures, this will be a source of bug and display the wrong thing. Not a concern for now though.

aryamanshodhan (approved):
Well done Sophia. Please wait for Cole to respond to this prior to merging in. In the meantime, please work on testing this PR thoroughly. We should not find bugs after the code is merged in. Therefore keep testing it.

csmanel (approved):

PR #156, opened 2026-06-30
bug-fix: add clear local storage on handle logout

csmanel:


aryamanshodhan (commented):
Looks good, but what happens if the user gets logged out due to inactivity?

PR #161, opened 2026-07-03
VCV-239: active devices should be based on collection cycles not months

csmanel:
### Summary
Re-bases the Operations map's **Devices** view from rolling calendar months onto **Collection Cycles**, so "active" means the same window as Review and DHIS2 submission. Also extracts a generic map info panel shared by the Devices and Specimens views.

### Changes
- **Activity model → cycles.** A device is **Active** if it has ≥1 session in the current cycle, **Inactive** if it's in the 3-cycle universe but silent this cycle. Membership is read from the backend `session.collectionCycleId` (id-equality), not re-derived on the frontend — no boundary-day timezone bug. Programs with no Collection Schedule fall back to a calendar-month window (mirrors `buildReviewSegments`).
- **Dropped "Lapsing."** Middle tier removed; marker color collapses to binary (active/inactive) across cards, legend, and info panel.
- **Generic `MarkerInfoPanel`.** Fed a `MarkerRow[]`; `SpecimenInfoPanel` and `DeviceInfoPanel` are thin feature adapters. Rows now group under their parent location path.

### Notes for review
- The month-window fallback is explicitly temporary — remove once all programs have cycles (see ADR-0006).
- Marker/headline totals are derived once in `buildDeviceMarkers` over the on-map (filtered) marker set.


scho324 (dismissed):

src/features/operations/geographical-summary/components/device-info-panel.tsx
```diff
@@ -20,39 +22,36 @@ export default function DeviceInfoPanel({
...
-                        key={marker.id}
-                        isSelected={isSelected}
-                        isLoading={isLoading}
-                        onSelect={() =>
-                            onMarkerSelect(isSelected ? null : marker.id)
-                        }
-                    >
-                        <p className="font-semibold">
-                            {marker.siteName ?? t('unknownSite')}
+    const markerRows = useMemo<MarkerRow[]>(
```
aryamanshodhan: Instead of casting to MarkerRow, is it possible to pass the markers directly and just display the elements inside the component itself? It seems MarkerRow[] is entirely derived from the markers, and then being displayed in the component. That way we can get rid of a derived intermediate field.

src/features/operations/geographical-summary/components/marker-info-panel.tsx
```diff
@@ -1,13 +1,19 @@
...
 import SelectableInfoPanel from '@/features/operations/geographical-summary/components/selectable-info-panel';
 import SelectableInfoPanelRow from '@/features/operations/geographical-summary/components/selectable-info-panel-row';
-import { useMemo } from 'react';
+
+export interface MarkerRow {
+    id: string;
+    siteName: string | null;
+    parentLocationName: string;
+    details: ReactNode;
+}
```
aryamanshodhan: This is not the appropriate place for this interface.

src/features/operations/geographical-summary/components/specimen-info-panel.tsx
```diff
@@ -0,0 +1,86 @@
...
+} from '@/features/operations/geographical-summary/components/marker-info-panel';
+
+interface SpecimenInfoPanelProps {
+    markers: SiteMarker[];
+    selectedMarkerId: string | null;
+    onMarkerSelect: (id: string | null) => void;
+    isLoading: boolean;
+}
+
+export default function SpecimenInfoPanel({
```
aryamanshodhan: Same as comment for devices. It seems MarkerRow is just an intermediate derived type and you might be able to get away with just getting rid of it if you display directly insisde the component.

aryamanshodhan (commented):
Will check for correctness after the comments on this PR are addressed. In addition to the comments left, I think the utility files all need a second pass through because I noticed a lot of intermediate types and functions that were added. Reassess whether they are absolutely needed or not, and if not then just remove them.

aryamanshodhan (changes requested):

aryamanshodhan (commented):
Just verify correct display on production prior to merging in. Test shows everything as inactive.

aryamanshodhan (approved):

PR #162, opened 2026-07-06
VCV-234: Show if total specimens on review page 2 matches expected amount

scho324:
* Reliable expected count exists if none of the sessions have their expectedSpecimens as undefined and if the total expected count >= the number of specimens uploaded 
* If there is a reliable expected count, displays '# Uploaded of # Expected Specimens'; otherwise would be redundant, since the 'Specimen # of #' already exists right below it 
* If there are specimens missing, displays an alert circle next to the text and makes the text red 

Update: 
* Red badge appears if there is a reliable expected count and there are specimens missing; hovering over it displays text about how many specimens are missing and tells the user to contact the relevant field user
* Passed down fetched sessions from the parent, image workspace no longer re-fetches them


src/features/review/workspace/components/review-workspace-page-client.tsx
```diff
@@ -137,6 +137,8 @@ export default function ReviewWorkspacePageClient({
                         ) : currentStepIndex === 1 ? (
                             <ImageReviewWorkspace
                                 siteId={siteId}
+                                siteName={siteName || 'Unknown'}
```
aryamanshodhan: I would not pass in siteName. Since we are now passing siteId and siteName, I would make it so yuo pass the entire site instead. And then you can maybe just display site name from there, or if it is not available, display the district/house number, etc.

src/features/review/workspace/image/components/image-review-workspace.tsx
```diff
@@ -70,6 +81,23 @@ export default function ImageReviewWorkspace({
...
+    const isMissingExpectedCounts = sessions.some(
+        session => session.expectedSpecimens === undefined,
+    );
+    const hasReliableExpectedCount =
+        !isMissingExpectedCounts &&
+        expectedSpecimensCount > 0 &&
+        totalSpecimensUploaded <= expectedSpecimensCount;
+    const isMissingExpectedSpecimens =
+        hasReliableExpectedCount &&
+        totalSpecimensUploaded < expectedSpecimensCount;
```
aryamanshodhan: I would recommend not displaying X specimens submitted of X expected. Because this expected specimens feature was released in Feb, so some sessions might have 0 expected specimens but there might actually be stuff uploaded. Therefore, I would just check to see if expected > actual, and only then display a warning saying "X images were not uploaded. Please contact the field user responsible for this site"

aryamanshodhan (changes requested):

src/features/review/workspace/image/components/image-review-workspace.tsx
scho324: I currently have it set to check if there's a reliable expected count (if all sessions have an expected number of specimens, and if uploaded <= total expected), and if there are specimens missing from a reliable expected count, to display the badge that has that message when hovered over.

src/features/review/workspace/components/review-workspace-page-client.tsx
```diff
@@ -136,7 +136,8 @@ export default function ReviewWorkspacePageClient({
                             )
                         ) : currentStepIndex === 1 ? (
                             <ImageReviewWorkspace
-                                siteId={siteId}
+                                site={site!}
```
csmanel, line 139: Will this properly error if site is actually undefined?

src/features/review/workspace/image/components/image-review-workspace.tsx
```diff
@@ -70,6 +80,23 @@ export default function ImageReviewWorkspace({
     }
 
     const specimens = getAllSpecimensResult.data.specimens;
+    const allSpecimensForSite = getAllSpecimensResult.data.specimens;
+
+    const totalSpecimensUploaded = allSpecimensForSite.length;
```
csmanel: Do we need this if we have `const totalSpecimens = specimens.length;` already?

src/features/review/workspace/image/components/image-review-workspace.tsx
```diff
@@ -70,6 +80,23 @@ export default function ImageReviewWorkspace({
...
+    const expectedSpecimensCount = sessions.reduce(
+        (total, session) => total + (session.expectedSpecimens ?? 0),
+        0,
+    );
+    const isMissingExpectedCounts = sessions.some(
+        session => session.expectedSpecimens === undefined,
+    );
+    const hasReliableExpectedCount =
```
csmanel: I am curious, do we need this? Could we collapse this all into a single check

could it be something more like 
```
const expectedSpecimenTotal = sessions.reduce(
    (total, session) => total + (session.expectedSpecimens ?? 0),
    0,
);
const missingSpecimenCount = expectedSpecimenTotal - specimens.length;
```

src/features/review/workspace/image/components/image-review-workspace.tsx
```diff
@@ -70,6 +80,23 @@ export default function ImageReviewWorkspace({
...
+        !isMissingExpectedCounts &&
+        expectedSpecimensCount > 0 &&
+        totalSpecimensUploaded <= expectedSpecimensCount;
+    const isMissingExpectedSpecimens =
+        hasReliableExpectedCount &&
+        totalSpecimensUploaded < expectedSpecimensCount;
 
     if (specimens.length === 0) {
```
csmanel, line 89: What happens in the scenario where there should be specimens, but we see none?

Will we see the missing specimens badge still?

src/features/review/workspace/image/components/image-review-workspace.tsx
```diff
@@ -100,12 +127,39 @@ export default function ImageReviewWorkspace({
...
+                                    {t('missingSpecimensBadge')}
+                                </Badge>
+                            </TooltipTrigger>
+                            <TooltipContent>
+                                {t('specimensNotUploaded', {
+                                    count:
+                                        expectedSpecimensCount -
+                                        totalSpecimensUploaded,
+                                    site:
+                                        site.name ??
```
csmanel: can we reuse `getSiteLabelParts` ?

src/features/review/workspace/image/components/image-review-workspace.tsx
scho324: Not sure, do we still want to display a warning of specimens missing if the expected count is unreliable but enough are missing that it still shows the badge? It'd correctly say that there are specimens missing but not a reliable count of how many are missing.

src/features/review/workspace/components/review-workspace-page-client.tsx
scho324: Will there ever be a case that the user cannot access all the sites for a program? Even if I am on the image review step and type in some other siteId in the URL outside of the ones in my permissions, remove the collection cycle parameter, and switch the date range to be broader, it goes back to the metadata review step and displays the 'No sessions found for this Sentinel Site' message.

scho324: The alternative is to go back to passing in siteId and siteName instead of site so that we don't need to worry about site being undefined

src/features/review/workspace/components/review-workspace-page-client.tsx
```diff
@@ -137,8 +137,9 @@ export default function ReviewWorkspacePageClient({
                         ) : currentStepIndex === 1 ? (
                             <ImageReviewWorkspace
                                 siteId={siteId}
-                                siteName={siteName || 'Unknown'}
+                                siteName={siteName}
                                 sessions={sessions}
+                                readOnly={readOnly}
```
aryamanshodhan: Not sure I understand the purpose of readOnly here.

src/features/review/workspace/components/review-workspace-page-client.tsx
```diff
@@ -137,8 +137,9 @@ export default function ReviewWorkspacePageClient({
                         ) : currentStepIndex === 1 ? (
                             <ImageReviewWorkspace
                                 siteId={siteId}
-                                siteName={siteName || 'Unknown'}
+                                siteName={siteName}
```
aryamanshodhan: Do we not have access to the full site to be able to pass it in? instead of siteId and siteName?

src/features/review/workspace/image/components/image-review-workspace.tsx
```diff
@@ -52,13 +54,13 @@ export default function ImageReviewWorkspace({
     const specimenQueryParams: GetAllSpecimensQueryParams =
         collectionCycleId !== undefined
             ? {
-                  siteId,
+                  siteId: siteId,
```
aryamanshodhan: This is not needed. When you say just siteId, it means the same thing, since the field and variable have the same name.

src/features/review/workspace/image/components/image-review-workspace.tsx
```diff
@@ -52,13 +54,13 @@ export default function ImageReviewWorkspace({
...
                   collectionCycleId,
                   sessionType: 'SURVEILLANCE',
                   includeAllImages: true,
               }
             : {
-                  siteId,
+                  siteId: siteId,
```
aryamanshodhan: same here

aryamanshodhan (dismissed):

src/features/review/workspace/components/review-workspace-page-client.tsx
scho324: I don't know if there's a point showing the missing-specimens alert if the user can't do anything about certified sessions. I can show it regardless if you'd like.

src/features/review/workspace/components/review-workspace-page-client.tsx
scho324: Cole was concerned about what to do if site was undefined so I changed it back; it only broke when I manually set site to null, but I'm not sure if there will ever be a case that the site with the relevant siteId stops existing and can't be fetched.

csmanel (approved):

csmanel (dismissed):
This is looking good! 

one optional thing if you would like to implement, the Tooltip is duplicated twice and likely could be abstracted.

https://github.com/vci-developers/vector-verify/pull/162/changes/BASE..6d88135573b8fbb8393cfcb1264707dc63df69ae#diff-d907d1a0d14a5c683d2ae219c219554b7978dce44b7652630795f2eb1777f81cR97

csmanel (approved):

PR #163, opened 2026-07-08
VCV-235: tracking user analytics

csmanel:
Adds a devMode-gated **User Analytics** dialog to the sidebar that charts rolling
active web-app-user counts (A1 / A7 / A30) over a selectable window (30d / 90d / 1y),
for a single program or all programs combined.

### Changes
- **BFF** — `GET /api/users/active-metrics/all` paginates the backend
  `GET /users/active-metrics` into one aggregated response (backend caps `limit`
  at 365, and the 1y window is an inclusive 366 days, so a full year needs the loop).
- **Data layer** — `getAllUserActiveMetrics` server fn, `useGetAllUserActiveMetrics`
  hook, query key, and Zod schemas validated against the live API spec.
- **UI** — `UserAnalyticsDialog` with a program selector, range tabs, and a
  Recharts line chart (`ActiveUserTrendChart`).


src/api/user/validation/get-all-user-active-metrics-schema.ts
```diff
@@ -0,0 +1,25 @@
+import { z } from 'zod';
+import { booleanQueryParamSchema } from '@/lib/network/validation/boolean-query-param-schema';
+import { activeMetricSnapshotSchema } from '@/api/user/validation/get-user-active-metrics-schema';
+
+export const getAllUserActiveMetricsQueryParamsSchema = z.object({
+    startDate: z.string().optional(),
+    endDate: z.string().optional(),
```
aryamanshodhan: I'd make startDate and endDate required for this queryParamSchema to avoid infinitely long lists. The user should be able to control the startDate and endDate

src/api/user/get-all-user-active-metrics.ts
```diff
@@ -0,0 +1,54 @@
...
+import {
+    type GetAllUserActiveMetricsQueryParams,
+    type GetAllUserActiveMetricsResponseBody,
+} from '@/api/user/validation/get-all-user-active-metrics-schema';
+import { ok, type Result } from '@/lib/result/result';
+import type { NetworkError } from '@/lib/network/network-error';
+import { constructQueryString } from '@/lib/network/construct-query-string';
+import { safeApiCall } from '@/lib/network/safe-api-call';
+
+const LIMIT = 365;
```
aryamanshodhan, line 16: Does backend allow a limit > 100? I know the answer to this is no for the other endpoints...not sure about this one.

src/api/user/validation/get-user-active-metrics-schema.ts
```diff
@@ -0,0 +1,42 @@
...
+export const getUserActiveMetricsQueryParamsSchema = z.object({
+    startDate: z.string().optional(),
+    endDate: z.string().optional(),
+    programId: z.coerce.number().optional(),
+    globalOnly: booleanQueryParamSchema().optional(),
+    limit: z.coerce.number().min(1).max(365).optional(),
+    offset: z.coerce.number().min(0).optional(),
+});
+
+export const activeMetricSnapshotSchema = z.object({
```
aryamanshodhan: I'd move this into its own file since it is being used in multiple places. You can make it a core domain model.

src/components/layout/app-sidebar.tsx
```diff
@@ -99,6 +102,8 @@ export default function AppSidebar({ userProfile }: AppSidebarProps) {
     } = useGetUserPermissions();
 
     const [isRawDataExportOpen, setIsRawDataExportOpen] = useState(false);
+    const [isUserAnalyticsOpen, setIsUserAnalyticsOpen] = useState(false);
+    const t = useTranslations('UserAnalytics');
```
aryamanshodhan, line 106: Flagging this for later, not now. But "Raw data export" is hard coded here, not in messages.json. We might need to rename t to something else, and might need to reoganize messages to get the raw strings for this page.

src/components/user-analytics/build-active-metrics-range.ts
aryamanshodhan, line 1: Perform a second pass through this file...I feel some stuff can be simplified. Especially the ActiveMetricsDateRange. I feel like it is not specific enough that you need to have an intermediate type for it. It only contains the startDate and endDate.

src/components/user-analytics/user-analytics-dialog.tsx
```diff
@@ -0,0 +1,125 @@
...
+    open: boolean;
+    onOpenChange: (open: boolean) => void;
+}
+
+export default function UserAnalyticsDialog({
+    open,
+    onOpenChange,
+}: UserAnalyticsDialogProps) {
+    const t = useTranslations('UserAnalytics');
+    const [selectedProgram, setSelectedProgram] = useState<number | 'all'>(
```
aryamanshodhan: The user should only be able to see this data for the program they are logged into. Don't mix up programs.

src/components/user-analytics/user-analytics-dialog.tsx
```diff
@@ -0,0 +1,125 @@
...
+                            yAxisLabel={t('axisActiveUsersLabel')}
+                        />
+                    )}
+                </div>
+            </DialogContent>
+        </Dialog>
+    );
+}
+
+function ActiveUserStateMessage({ message }: { message: string }) {
```
aryamanshodhan: One component per file

src/components/user-analytics/user-analytics-program-select.tsx
```diff
@@ -0,0 +1,68 @@
...
+    isLoading: boolean;
+    hasError: boolean;
+}
+
+export default function UserAnalyticsProgramSelect({
+    value,
+    onValueChange,
+    programs,
+    isLoading,
+    hasError,
```
aryamanshodhan: Instead of isLoading and hasError, should we fetch using the hook here?

src/components/user-analytics/user-analytics-range-tabs.tsx
```diff
@@ -0,0 +1,45 @@
...
+import { Tabs, TabsList, TabsTrigger } from '@/components/ui/tabs';
+import type { ActiveMetricsRangePreset } from '@/components/user-analytics/build-active-metrics-range';
+import { useTranslations } from 'next-intl';
+
+interface UserAnalyticsRangeTabsProps {
+    value: ActiveMetricsRangePreset;
+    onValueChange: (value: ActiveMetricsRangePreset) => void;
+}
+
+const RANGE_PRESETS: { preset: ActiveMetricsRangePreset; labelKey: string }[] =
```
aryamanshodhan: RANGE_PRESET_TABS maybe?

aryamanshodhan (changes requested):
Checking for code cleanliness. Once the comments are addressed, I will look into correctness. Please address the comments, and also another pass through to ensure there are no unnecessary intermediate types/functions.

src/api/user/get-all-user-active-metrics.ts
csmanel: This endpoint is the exception, caps limit at 365 (minimum: 1, maximum: 365, default: 30), and the full 1y window needs it since 366 inclusive days spans two pages.

aryamanshodhan (approved):
Excellent work Cole! You managed to shave off over 150 lines of code. Please double check correctness prior to merging it in. The chart looks correct to me, but make sure it accurately reflects what was returned from the backend.

PR #164, opened 2026-07-08
VCV-250: Make all metadata fields editable

aryamanshodhan:


scho324 (commented):
Spacing between the 'Conflicting values must be resolved to continue' and 'Edit any value to correct it (optional)' notes seems to be off; when they're in the same line, the pencil icon is too close to the previous text. Everything else seems fine!

csmanel (dismissed):

scho324 (dismissed):

aryamanshodhan (commented):
Looks good to me. Please double check correctness across Uganda and non-uganda programs. For both conflicted and non-conflicted fields. And dev and non-dev accounts of course.

csmanel (dismissed):

scho324 (approved):

PR #165, opened 2026-07-09
VCV-245: Add a legend for species composition

scho324:
* Legend appears at the bottom of each chart card

### DIFF
--- a/src/features/operations/specimen-composition/components/composition-chart-pair.tsx
+++ b/src/features/operations/specimen-composition/components/composition-chart-pair.tsx
@@ -190,6 +190,26 @@ export default function CompositionChartPair({
                         </ChartContainer>
                     </div>
                 </div>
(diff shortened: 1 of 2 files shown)


src/features/operations/specimen-composition/components/composition-chart-pair.tsx
aryamanshodhan, line 1: I think Legend is already an existing component in the shadcn chart component. Check it out and use that instead of building your own legend.

aryamanshodhan (changes requested):

src/features/operations/specimen-composition/components/composition-chart-pair.tsx
scho324: Just switched to the legend in the chart component

csmanel (approved):

aryamanshodhan (approved):

PR #170, opened 2026-07-14
VCV-268: Display session timestamps along with the date in the site review so that supervisors can tell repeat visits apart

scho324:
* Session headers in the metadata table are now 'MMM d, yyyy  h:mm a'


src/features/review/workspace/metadata/components/metadata-review-table.tsx
```diff
@@ -68,6 +68,11 @@ export default function MetadataReviewTable({
                                     session.collectionDate,
                                     timezone,
                                     'MMM d, yyyy',
+                                )}{' '}
+                                {formatDateInTimezone(
+                                    session.createdAt,
+                                    timezone,
+                                    'h:mm a',
```
aryamanshodhan: This should also show the dd mm yyyy because it might not be the same as the collection date.

aryamanshodhan (changes requested):

src/features/review/workspace/metadata/components/metadata-review-table.tsx
scho324: As far as I've seen, the collection timestamp and the createdAt timestamp seem to be the same down to the second. Should I still add the createdAt date?

src/features/review/workspace/metadata/components/metadata-review-table.tsx
```diff
@@ -68,6 +68,11 @@ export default function MetadataReviewTable({
                                     session.collectionDate,
                                     timezone,
                                     'MMM d, yyyy',
+                                )}{' '}
+                                {formatDateInTimezone(
+                                    session.createdAt,
```
csmanel: I think we need to edit this just slightly so it is more clear 
We should add labels so its evident what the dates belong to, and so its more clear that the session creation date doesn't necessarily align with the collection date

```
│ Collection Date: Jul 10 │ 
│ Session Created:        │   
│   Jul 13, 2026 3:14 PM  │

```

csmanel (changes requested):

csmanel (approved):

aryamanshodhan (approved):

PR #172, opened 2026-07-16
VCV-248: Add a percent-confidence to AI prediction

scho324:
* Added an additional section in the image review section for model confidence
* While loading or if there is no prediction for a category, the percent confidence will show up as a dash 
* If the percent confidence is below 75%, it will be red and have a triangle alert next to it

### DIFF
--- a/messages/en.json
+++ b/messages/en.json
@@ -156,7 +156,8 @@
         "modelPredictions": "Model Predictions",
         "species": "Species",
         "sex": "Sex",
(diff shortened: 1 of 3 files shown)


aryamanshodhan (commented):
Code is fine. Stylistic question: Should we display the confidence percentages next to the model output instead of making it a separate section?

csmanel (approved):

PR #175, opened 2026-07-22
VCV-247: Switch line graphs to bar graphs

scho324:
* Toggle is in the top right, next to the species filter
* Toggle is set to bar graphs by default 
* Moved the line and bar charts to their own components so the pie chart doesn't refresh when the chart type changes


src/features/operations/specimen-composition/components/operations-specimen-composition.tsx
```diff
@@ -93,31 +99,57 @@ export default function OperationsSpecimenComposition({
...
-                    }
-                >
-                    <MultiSelectTrigger>
-                        <MultiSelectValue placeholder={t('selectSpecies')} />
-                    </MultiSelectTrigger>
-                    <MultiSelectContent
-                        search={{
-                            placeholder: t('searchSpecies'),
-                            emptyMessage: t('noSpeciesFound'),
+                <div className="flex-gap flex justify-between">
```
csmanel: I think there should be a space between `flex` and `gap`?

src/features/operations/specimen-composition/components/operations-specimen-composition.tsx
csmanel, line 1: Double check that you are using i18n when possible

src/components/ui/toggle-group.tsx
```diff
@@ -0,0 +1,64 @@
+'use client';
+
+import * as React from 'react';
+import { cva, type VariantProps } from 'class-variance-authority';
```
csmanel: Check out this lint warning

src/features/operations/specimen-composition/components/composition-bar-chart.tsx
csmanel, line 1: I think the two charts are fairly similar, do they need to be two separate components? Curious if there is an opportunity here to dry it up a bit

csmanel (changes requested):
The graph view helps a ton, solid work! Just some very minor changes

src/components/ui/toggle-group.tsx
aryamanshodhan, line 1: Question: Was this a component fetched directly from shadcn?

aryamanshodhan (dismissed):
Code is clean. Please double check for correctness. Also have a quick question on a component that was added.

src/components/ui/toggle-group.tsx
scho324: No, the original fetched component didn't add any spacing between the toggle buttons when I specified a nonzero spacing nor would it fuse the buttons if the spacing was 0. I'm not sure if it's an issue with my laptop's viewing since it also displayed improper spacing between some notes in VCV-250 even though Cole told me it looked fine on his end.

csmanel (dismissed):

csmanel (approved):

aryamanshodhan (approved):

PR #177, opened 2026-07-23
VCV-292: Give nominatimSearch country param so markers are more accurate

csmanel:
## Summary

Site markers on the Operations map can render in the wrong country entirely (e.g. a Ghana site landing in Sri Lanka), because Nominatim disambiguates same-named regions by importance alone. 
This PR makes every geocode lookup toward the site's actual program country via Nominatim's `countrycodes` param.

- Added `i18n-iso-countries` to resolve the program's stored country name (e.g. `"Ghana"`) to an ISO alpha-2 code (`gh`) that Nominatim's `countrycodes` expects. Avoids hand-maintaining a name→code map that would need updating on every new program country; the library's static table covers every country up front. 
- `country` is now a required input threaded through `SiteMap` and the device-activity map (`DeviceView`/`DeviceMap`) down to the shared geocoding hook, so no marker geocodes without a country bias.

### DIFF
--- a/src/features/operations/geographical-summary/components/device-map.tsx
+++ b/src/features/operations/geographical-summary/components/device-map.tsx
@@ -10,6 +10,7 @@ import MarkerInfoPanel from '@/features/operations/geographical-summary/componen
 
 interface DeviceMapProps {
     markers: DeviceMarker[];
+    country: string;
     selectedLocations: string[];
     selectedMarkerId: string | null;
     onMarkerSelect: (id: string | null) => void;
@@ -17,6 +18,7 @@ interface DeviceMapProps {
 
 export default function DeviceMap({
     markers,
+    country,
     selectedLocations,
     selectedMarkerId,
     onMarkerSelect,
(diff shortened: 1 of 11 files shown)


scho324 (commented):
Will the map navigator also need the country?

scho324 (approved):

src/features/operations/geographical-summary/components/device-map.tsx
```diff
@@ -31,6 +33,7 @@ export default function DeviceMap({
     return (
         <GeocodedClusterMap
             markers={markers}
+            country={country}
```
aryamanshodhan, line 36: Question: Should we make country part of the marker datatype?

src/features/operations/geographical-summary/components/operations-geographical-summary.tsx
```diff
@@ -39,15 +40,36 @@ export default function OperationsGeographicalSummary({
...
         endDate,
     });
 
+    const {
+        data: getProgramsResult,
+        isPending: isProgramsPending,
+        isError: isProgramsError,
+    } = useGetPrograms();
+
+    const country = getProgramsResult?.ok
```
aryamanshodhan, line 61: Note for later. We should consider caching and context/state management better so these few computations can be one-time cached entries that we dont need to repeatedly calculate and prop-drill.

aryamanshodhan (approved):
Submitted an optional comment. Please look into it. Otherwise, code looks clean. Please test for correctness.

PR #179, opened 2026-07-27
VCV-284: Create a panel to move existing assigned sessions to a different collection cycle

csmanel:
## Summary

Adds a Sessions tab to the Review page so supervisors can find a session and move it to a different collection cycle, without needing to dig through the Sites List hierarchy.

- New Sessions tab: search, sort, filter by cycle/review state, and reassign a session's collection cycle from one table. Scoped to `type: SURVEILLANCE`
- Reassigning a session into or out of a cycle that's already fully reviewed (`SUBMITTED` or `CERTIFIED`) resets the affected sessions in that cycle back to `NEEDS_REVIEW`, with a confirmation dialog before that happens
- Sessions tab supports an "All locations" option in the location picker, since reassignment isn't scoped to a single site the way the other tabs are
- Header restructured so tabs and filters live in separate rows, and the filter set now adapts per tab (e.g. review state defaults differently entering vs. leaving the Sessions tab)
- `PUT /sessions/:id` now accepts `collectionCycleId`, with a refine requiring at least one of the two
- `SessionsTableRow` is memoized with stabilized callbacks to reduce re-renders while searching. Helps, but doesn't fully solve it at scale — the table isn't virtualized, so all rows still mount at once. Worth revisiting with row virtualization if session counts keep growing.

Some links regarding table row memoization and virtualization 
- https://www.npmjs.com/package/react-window
- https://github.com/TanStack/virtual/discussions/535


src/features/review/sessions-table/components/sessions-table-row.tsx
```diff
@@ -0,0 +1,123 @@
...
+    SelectTrigger,
+    SelectValue,
+} from '@/components/ui/select';
+import { TableCell, TableRow } from '@/components/ui/table';
+import { formatDateInTimezone } from '@/utils/format-date-in-timezone';
+import Link from 'next/link';
+import { useTranslations } from 'next-intl';
+import { memo } from 'react';
+
+const SESSION_STATE_LABEL_KEY: Record<SessionState, string> = {
```
csmanel: can we use REVIEW STATE LABEL KEY?

src/features/review/sessions-table/components/sessions-table-row.tsx
```diff
@@ -0,0 +1,123 @@
...
+    ...REVIEW_STATE_LABEL_KEY,
+    NOT_APPLICABLE: 'notApplicable',
+};
+
+interface SessionsTableRowProps {
+    session: Session;
+    site: Site | undefined;
+    cycle: CollectionCycle | undefined;
```
csmanel: differentiate between cycle and collectioncycle[] (naming doesn't say enough

src/features/review/sessions-table/components/sessions-table.tsx
```diff
@@ -0,0 +1,287 @@
...
+    }
+
+    if (rowData.length === 0) {
+        return (
+            <div className="flex flex-col space-y-4">
+                {searchInput}
+                <p className="text-muted-foreground py-12 text-center text-sm">
+                    {t('noSessions')}
+                </p>
+                {reassignmentDialog}
```
scho324: Do we need the reassignment dialog if there's no results?

scho324 (commented):
If I start from default filter settings and then switch the date range from May-July to May-May, sessions that belong to another cycle but were collected/created in May are displayed as unassigned, not sure if this is intended behavior but it might be confusing to see it change and it might have unintended consequences. Also, the placeholder text for an unassigned session isn't showing up for me, but it might just be my computer. 

Maybe it would be good to define what a 'Review Unit' is to users in the confirmation dialog?

Otherwise the code looks okay; I'm not sure where it would be reduced since it feels like everything is needed for readability.

src/api/session/hooks/use-get-all-sessions.ts
```diff
@@ -19,7 +19,7 @@ type GetAllSessionsQueryOptions = Omit<
     'queryKey' | 'queryFn'
 >;
 
-async function fetchAllSessions(
+export async function fetchAllSessions(
```
aryamanshodhan: This is the client function. We should use the query hook instead of consuming the client function directly.

src/api/session/validation/put-session-by-id-schema.ts
```diff
@@ -4,9 +4,17 @@ import {
     sessionStateSchema,
 } from '@/api/session/validation/session-schema';
 
-export const putSessionByIdRequestSchema = z.object({
-    state: sessionStateSchema,
-});
+export const putSessionByIdRequestSchema = z
```
aryamanshodhan, line 7: You can just use .partial() instead of specifying optional for both fields.

src/components/ui/table.tsx
```diff
@@ -3,11 +3,18 @@
 import * as React from 'react';
 import { cn } from '@/utils/cn';
 
-function Table({ className, ...props }: React.ComponentProps<'table'>) {
+function Table({
+    className,
+    containerClassName,
```
aryamanshodhan, line 8: Difference between className and containerClassName?

src/features/review/sessions-table/components/sessions-table-row.tsx
```diff
@@ -0,0 +1,117 @@
...
+import { TableCell, TableRow } from '@/components/ui/table';
+import { formatDateInTimezone } from '@/utils/format-date-in-timezone';
+import Link from 'next/link';
+import { useTranslations } from 'next-intl';
+import { memo } from 'react';
+
+interface SessionsTableRowProps {
+    session: Session;
+    site: Site | undefined;
+    sessionCollectionCycle: CollectionCycle | undefined;
```
aryamanshodhan, line 24: This is fine for now, but we really need context.

src/features/review/sessions-table/components/sessions-table.tsx
```diff
@@ -0,0 +1,293 @@
...
+    sites: Site[];
+    collectionCycles: CollectionCycle[];
+    selectedCycleIds: number[];
+    selectedReviewStates: SessionState[];
+    locationQueryParam: LocationQueryParam | undefined;
+    startMonth: Date;
+    endMonth: Date;
+}
+
+function matchesSearch(session: Session, siteName: string, search: string) {
```
aryamanshodhan: Util function

src/features/review/sessions-table/components/sessions-table.tsx
```diff
@@ -0,0 +1,293 @@
...
+                session =>
+                    selectedReviewStates.length === 0 ||
+                    (session.state !== undefined &&
+                        selectedReviewStates.includes(session.state)),
+            )
+            .filter(session => {
+                const siteName =
+                    getSiteDisplayName(siteById.get(session.siteId)) ?? '';
+                return matchesSearch(session, siteName, search);
+            });
```
aryamanshodhan: Why do we need filter after filter after filter? Why not just do everything within one?

src/features/review/sites-list/components/review-sites-list-page-client.tsx
```diff
@@ -89,6 +92,20 @@ export default function ReviewSitesListPageClient() {
         : [];
 
     function handleTabChange(tab: ReviewTab) {
+        const isSessionsTab = tab === 'sessions' || activeTab === 'sessions';
+
+        if (isSessionsTab) {
```
aryamanshodhan, line 97: Just confirming...filters for sessions tab should work the same as the other tabs for now.

src/features/review/sessions-table/utils/sort-sessions.ts
aryamanshodhan, line 1: See if this file can be cleaned up.

src/features/review/sessions-table/utils/get-site-display-name.ts
aryamanshodhan, line 1: Can we promote this to shared? Is this something that is useful for operations or annotations as well?

src/features/review/sessions-table/hooks/use-reassign-session-cycle.ts
aryamanshodhan, line 1: Consider cleanup of this file. See if things can be simplified (Eg. intermediate types, reduce number of util functions, etc.)

aryamanshodhan (changes requested):
Just making sure, sessions tab should have the start date end date filters, and collection cycle filters working identically to the other tabs within the review feature - submissions and review sites list.

src/components/ui/table.tsx
csmanel: className styles the table element itself
containerClassName styles the wrapper 

It is needed it for SessionsTable's sticky header + max-height scroll container (get that effect of the table is contained to a scrolling container/wrapper rather than it splaying out on the page)

src/features/review/sites-list/components/review-sites-list-page-client.tsx
csmanel: yes, sessions tab uses the same startMonth/endMonth → startDate/endDate and selectedCycleIds mechanism as the sites-list and submissions tabs

The one intentional difference is selectedReviewStates defaulting to [] instead of ['NEEDS_REVIEW'] on this tab

src/features/review/sessions-table/utils/sort-sessions.ts
csmanel: I tried to clean it up but I couldn't exactly find good things to remove/simplify.
I considered bundling the siteById/cycleById map params into a single lookup object, but that would just be an intermediate type. 
Outside of that the file seems fairly lean to me? If you have any specific things you would want changed I am happy to look into it.

src/features/review/sessions-table/utils/get-site-display-name.ts
csmanel: Operations (geographical-summary-helpers.ts) and dhis2-sync (get-site-label-parts.ts) already have their own site-label logic, so I'm not sure I'd promote this to shared right now?

Could be worth exploring in another PR whether the utils can be reworked to be usable across the board.

src/features/review/sessions-table/hooks/use-reassign-session-cycle.ts
csmanel: Tried to reduce the util functions and types here, but I am not sure what else can be done?

CycleReassignment is the one intermediate type, it's shared between the pending-confirmation state and the confirm action, so I think it makes sense to keep? 

Let me know if you had a specific piece in mind

aryamanshodhan (approved):

PR #180, opened 2026-07-30
VCV-275: Error state - Annotations

scho324:
* GET errors display error banner and red skeletons
--> Annotation form buttons and fields are disabled
--> Edit button on view-only disappears 
--> Tasks list just has the error banner, since there is no special structure to follow like the form/image view/view-only components do
--> Buttons to move between specimens only appear if fetching the annotations was successful

image

image

* PUT error in updating annotations shows a toast

image


src/features/annotation/task-details/components/workspace/annotation-workspace.tsx
```diff
@@ -67,13 +77,14 @@ export default function AnnotationWorkspace({
...
-    const annotation = getAnnotationsResult?.data.annotations[0];
-    const total = getAnnotationsResult?.data.total;
+    const annotation = getAnnotationsResult?.ok
+        ? getAnnotationsResult.data.annotations[0]
+        : undefined;
+    const total = getAnnotationsResult?.ok
+        ? getAnnotationsResult.data.total
+        : undefined;
 
     if (!isLoading && (!annotation || total === 0)) {
```
csmanel: I believe this conditional doesn't actually work for checking an empty non error state?

These should be separate cases and I am not sure that they currently are
- empty: the request succeeded, there's just nothing left to annotate -> should show "No annotations found"
- failed: the request errored -> should show the error banner

Double check that we are handling both true empty state and failed state, if not I think the fix is simply also checking that it didn't error. There may be a cleaner way of doing this though (if it is even needed)

```
const isEmpty = !isLoading && !isError && (!annotation || total === 0);
if (isEmpty) {
```

src/features/annotation/task-details/components/workspace/specimen-image-viewer.tsx
```diff
@@ -6,26 +6,36 @@ import { Card, CardContent, CardHeader, CardTitle } from '@/components/ui/card';
 import { Separator } from '@/components/ui/separator';
 import { Skeleton } from '@/components/ui/skeleton';
 import { MapPin } from 'lucide-react';
+import { useTranslations } from 'next-intl';
 import { Fragment } from 'react';
 
 interface SpecimenImageViewerProps {
-    specimen?: Specimen;
+    specimen?: Specimen | undefined;
```
csmanel: do we need undefined here?

csmanel (approved):

PR #182, opened 2026-08-06
VCV-271: Loading State - Operations

scho324:
* Geographical summary has the same loading state as before; updated devices to match specimens loading structure
* Specimen composition loads the three skeleton cards with a circle for the pie chart and a rectangle for the line/bar chart; please check if the rectangle shows up for h-full, because it disappears for anything larger than h-64 and I'm not sure if it's just my laptop
* AI performance has skeletons for missing information (coverage, reviewed specimens, accuracy) and skeletonLists for the table contents (confusion matrix, sensitivity/specificity) and displays any static content (interpretation)
* Field team performance is the same as before, except a small skeleton at the top was removed as it didn't fit with the card appearance

### DIFF
--- a/src/features/operations/ai-performance/components/operations-ai-performance.tsx
+++ b/src/features/operations/ai-performance/components/operations-ai-performance.tsx
@@ -5,6 +5,8 @@ import type { GetAnnotationsSummaryQueryParams } from '@/api/annotation/validati
 import { Card, CardContent } from '@/components/ui/card';
 import SpecimenConfusionMatrix from './specimen-confusion-matrix';
 import { Info } from 'lucide-react';
+import { Fragment } from 'react/jsx-runtime';
+import { Skeleton } from '@/components/ui/skeleton';
 
 interface OperationsAiPerformanceProps {
     siteIds: number[];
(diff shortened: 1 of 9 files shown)


src/features/operations/specimen-composition/components/composition-chart-pair.tsx
csmanel: totalSpecimenCount === 0 isn't gated on isLoading, so it shows "No specimen data" rather than skeletons during loading 

specimen-confusion-matrix.tsx has the same check I believe it displays skeletons during loading. Double check that and copy its pattern if it looks right

src/features/operations/geographical-summary/components/device-view.tsx
csmanel, line 1: We should still have some empty state view for the map in the scenario in which there is no data

Currently the map just loads with nothing on it, but I don't think that is particularly suggestive of "we have nothing to display on the map"

src/features/operations/ai-performance/components/operations-ai-performance.tsx
```diff
@@ -41,93 +44,122 @@ export default function OperationsAiPerformance({
...
-    const flaggedSpecimens = annotationsSummary.statusCounts.FLAGGED;
-    const reviewedSpecimens = annotatedSpecimens + flaggedSpecimens;
-    const totalSpecimens = annotationsSummary.total;
+    const annotationsSummary = getAnnotationsSummaryResult?.ok
+        ? getAnnotationsSummaryResult.data
+        : undefined;
+    const annotatedSpecimens = annotationsSummary?.statusCounts.ANNOTATED;
+    const flaggedSpecimens = annotationsSummary?.statusCounts.FLAGGED;
+    const reviewedSpecimens =
+        annotatedSpecimens && flaggedSpecimens
```
csmanel: Is this the same conditional as before? I think something potentially broke in the logic 

Showing there is one reviewed specimen, but the number is still 0

Image

csmanel (approved):
I think this looks good! Couldn't see any issues, do a final round of testing and then should be good to go!

PR #183, opened 2026-08-06
VCV-297: add empty state for species charts, fix species filter width

csmanel:
- Fix issue with width on species filter
- Add empty state for species composition graphs

### DIFF
--- a/messages/en.json
+++ b/messages/en.json
@@ -74,7 +74,8 @@
         "searchSpecies": "Search species...",
         "noSpeciesFound": "No species found.",
         "barChart": "Bar chart",
-        "lineChart": "Line chart"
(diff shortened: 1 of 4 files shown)


scho324 (dismissed):
If I have the site at full width and add too many species options from the filter (set date range to Jan 2025 to Aug 2026), it just widens the entire display to fit the extra ones on the same line until some larger maximum width where it decides to start adding the (+N) badge. 
If I have it at half width, it stays within the original bounds and adds the (+N) badge in the filter. If it's just me or this is intended/preferred over other alternatives like setting overflowBehavior, go ahead and merge; everything else looks okay.

scho324 (approved):

PR #186, opened 2026-08-14
VCV-274: Error State - Operations

scho324:
* Geographical summary: added error banner appearance for if there's an error getting the country; adjusted device-view formatting to match structure for specimen-view
* Specimen composition: add error banner, disable filter + bar/line graph toggle, make loading skeletons destructive on error 
* AI performance: make loading skeletons destructive on error; replace the 'coverage reflects' banner at the top with the error banner
* Field user compliance: add error banner, make loading skeleton destructive

Currently debating whether to match error states closer to geographical summary or to match geographical summary's error states to the other ones

### DIFF
--- a/src/features/operations/ai-performance/components/operations-ai-performance.tsx
+++ b/src/features/operations/ai-performance/components/operations-ai-performance.tsx
@@ -9,6 +9,7 @@ import { Fragment } from 'react/jsx-runtime';
 import { Skeleton } from '@/components/ui/skeleton';
 import { useTranslations } from 'next-intl';
 import ErrorBanner from '@/components/ui/error-banner';
+import { cn } from '@/utils/cn';
 
 interface OperationsAiPerformanceProps {
     siteIds: number[];
@@ -66,7 +67,12 @@ export default function OperationsAiPerformance({
 
             <div className="grid gap-3 lg:col-span-2 lg:grid-cols-2">
                 <Fragment>
-                    <Card className="border-success/40 bg-success/5 gap-0 py-0">
+                    <Card
+                        className={cn(
+                            !isError && 'border-success/40 bg-success/5',
+                            'gap-0 py-0',
+                        )}
+                    >
                         <CardContent className="p-4">
                             <p className="text-muted-foreground text-sm">
                                 {t('coverage')}
--- a/src/features/operations/ai-performance/components/operations-ai-performance.tsx
+++ b/src/features/operations/ai-performance/components/operations-ai-performance.tsx
@@ -6,10 +6,10 @@ import { Card, CardContent } from '@/components/ui/card';
 import SpecimenConfusionMatrix from './specimen-confusion-matrix';
 import { Info } from 'lucide-react';
 import { Fragment } from 'react/jsx-runtime';
-import { Skeleton } from '@/components/ui/skeleton';
 import { useTranslations } from 'next-intl';
 import ErrorBanner from '@/components/ui/error-banner';
 import { cn } from '@/utils/cn';
+import { StatCardSkeleton } from '@/components/ui/stat-card-skeleton';
 
 interface OperationsAiPerformanceProps {
     siteIds: number[];
@@ -78,22 +78,11 @@ export default function OperationsAiPerformance({
                                 {t('coverage')}
                             </p>
                             {isLoading || isError ? (
-                                <div className="space-y-2 py-2">
-                                    <Skeleton
-                                        width="sm"
-                                        height="xl"
-                                        variant={
-                                            isError ? 'destructive' : 'default'
-                                        }
-                                    />
-                                    <Skeleton
-                                        width="lg"
-                                        height="sm"
-                                        variant={
-                                            isError ? 'destructive' : 'default'
-                                        }
-                                    />
-                                </div>
+                                <StatCardSkeleton
+                                    variant={
+                                        isError ? 'destructive' : 'default'
+                                    }
+                                />
                             ) : (
                                 <Fragment>
                                     <p className="mt-1 text-4xl font-semibold tracking-tight">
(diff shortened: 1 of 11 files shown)


src/features/operations/ai-performance/components/operations-ai-performance.tsx
csmanel: The green is harcoded here, but if there is an error, the color of the card should reflect that. It looks a bit odd to have a green background/border when there is an error

src/features/operations/specimen-composition/components/operations-specimen-composition.tsx
csmanel: I think this has a similar issue to what we have discussed before?

We potentially run into issues with error state and empty state looking the same? 

At the moment I am seeing the empty state rather than error skeletons

src/features/operations/ai-performance/components/specimen-confusion-matrix.tsx
csmanel: I think potentially the same issue here where we get empty states rather than error states?

csmanel: I believe this is still partially an issue? 
I think these are the empty state messages not error state messages? Unless this is working as you intended 

Screenshot 2026-08-19 at 4 04 25 PM

(I was able to get this just by deleting my stored cookie)

src/features/operations/ai-performance/components/operations-ai-performance.tsx
csmanel: Fixed this, but I believe the error states aren't showing as intended? 

I think these are the empty state messages not error state messages?

Screenshot 2026-08-19 at 4 04 25 PM

scho324: Sorry, thought the error state exclusion was committed

src/features/operations/ai-performance/components/operations-ai-performance.tsx
```diff
@@ -62,22 +56,43 @@ export default function OperationsAiPerformance({
...
+                            'gap-0 py-0',
+                        )}
+                    >
                         <CardContent className="p-4">
                             <p className="text-muted-foreground text-sm">
                                 {t('coverage')}
                             </p>
-                            {isLoading ? (
+                            {isLoading || isError ? (
                                 <div className="space-y-2 py-2">
```
csmanel: In all of these places where we have a stat card skeleton, could we create a small component like SkeletonList that renders a SkeletonStatCard or something like that?

src/features/operations/geographical-summary/components/device-view.tsx
```diff
@@ -57,16 +58,6 @@ export default function DeviceView({
         );
     }, [deviceActivity, descendantsOfSelectedLocations, country]);
 
-    if (isError) {
```
csmanel: Now that we have gotten rid of the early return (which is the right call), on error, do we need to handle the hiding of the map legend so it doesn't display?

src/features/operations/geographical-summary/components/operations-geographical-summary.tsx
csmanel: Should this match our other conventions? 

`const isError = !isLoading && !getAnnotationsSummaryResult.ok;
`

Or is there a reason to keep it unchanged?

src/features/operations/geographical-summary/components/specimens-view.tsx
csmanel: I think this needs to check !isError as well

src/features/operations/intervention-metrics/components/operations-intervention-metrics.tsx
```diff
@@ -42,17 +44,18 @@ export default function OperationsInterventionMetrics({
     const failedSessionsMetricsQuery = sessionsMetricsQueries.find(
```
csmanel: can we change this to a .filter? 
That way if multiple districts fail all of them are named? 

Make sure that also works in the scenario that all fail, we don't then need to name every district.

src/features/operations/intervention-metrics/components/operations-intervention-metrics.tsx
```diff
@@ -42,17 +44,18 @@ export default function OperationsInterventionMetrics({
...
         failedSessionsMetricsQuery?.data &&
         !failedSessionsMetricsQuery.data.ok
-    ) {
-        return (
-            <p className="text-destructive text-sm">
-                {failedSessionsMetricsQuery.data.error.message}
-            </p>
-        );
-    }
+            ? (failedSessionsMetricsQuery.data.error.message ??
```
csmanel: Should we be consistent and prefer `t('failedToLoad')`

I believe `src/features/operations/field-user-compliance/components/operations-field-user-compliance.tsx#L56` sets this precedent, and I think the backend message isn't particularly helpful for a user?

I would be curious for your thoughts

src/features/operations/specimen-composition/components/operations-specimen-composition.tsx
```diff
@@ -167,10 +163,11 @@ export default function OperationsSpecimenComposition({
                             specimenCountsByMonth={[]}
                             specimenChartConfig={{}}
                             isLoading={isLoading}
+                            isError={isError}
                         />
                     ),
                 )
-            ) : validSelectedSpecies.length === 0 ? (
+            ) : !isError && validSelectedSpecies.length === 0 ? (
```
csmanel: I think because line 155 has both isLoading and isError, this !isError isn't actually necessary?

src/features/operations/intervention-metrics/components/operations-intervention-metrics.tsx
scho324: If the user selects a lot of districts and many of them fail but not all of them, wouldn't it still make us name a long list of them? I remember this was also discussed for the info banner in the AI performance tab when I was switching the location dropdown to a multiselect.

csmanel: Sorry I don't remember what was discussed, if we agreed that we shouldn't display all the names then thats fine

I just assumed it would make sense to display those that failed so the user isn't confused

src/features/operations/geographical-summary/components/device-view.tsx
scho324: Would it also look better to replace the AI performance interpretation with a skeleton for consistency?

csmanel: We likely should have stated this earlier on, but I think we should use the rule of 
Derived content, things that can load, should have skeletons, while static titles/headers/legends should not.

csmanel:
I think it looks good, do a final round of testing and then it should be good to go!

csmanel (approved):
all good to go! as always, final testing and add it to the excel!

PR #187, opened 2026-08-14
VCV-253: enforce required metadata fields, contrain select fields to a closed dropdown

csmanel:
- Flag rows with a real `required` signal (FormQuestion.required or
  SURVEILLANCE_FIELD_DEPENDENCIES dependents) blocks Save & Continue
- Constrain `select`-type fields to their real defined options via a closed
  dropdown, matching the existing `boolean` pattern; boolean itself no longer
  leaks arbitrary recorded values into its own dropdown
- Fix `resolveRowEffectiveValue` using raw `String(value)` instead of
  `formatDisplayValue`, which silently broke the disable for
  boolean such as (`wasIrsConducted`/`monthsSinceIrs`)

- Make the metadata review table scroll within its own container with a
  sticky header

### DIFF
--- a/messages/en.json
+++ b/messages/en.json
@@ -184,8 +184,10 @@
     "ReviewMetadata": {
         "intro": "Review the metadata for this unit before continuing.",
         "legendConflict": "Conflicting values must be resolved to continue",
+        "legendRequired": "Required answers must be supplied to continue",
         "legendEdit": "Edit any value to correct it (optional)",
         "conflictsRemaining": "{count, plural, one {# conflict left to resolve} other {# conflicts left to resolve}}",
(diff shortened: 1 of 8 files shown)


scho324 (commented):
For the Ghana account, the EJS-HLC-02 site under Ashanti cycle 4 has a 'false' recorded under 'Are domestic animals present in the house' instead of displaying 'No' like the dropdown options

csmanel:
> For the Ghana account, the EJS-HLC-02 site under Ashanti cycle 4 has a 'false' recorded under 'Are domestic animals present in the house' instead of displaying 'No' like the dropdown options

I will take a look, thanks!

scho324 (dismissed):

scho324 (approved):

PR #189, opened 2026-08-21
Docs/VCV-300: update docsite

csmanel:
update docsite to contain all recent changes and reflect the develop environment


public/docs/operations/02-specimen-composition.png
scho324: It seems like the time series looks empty?

csmanel: yeah I will fix that screenshot , thanks!

public/docs/review-detail/surveillance-questions-real.png
scho324: Not sure if this is used anywhere (quick find didn't have it mentioned anywhere), but it's similar to surveillance-edit-any-row.png

scho324 (approved):

PR #191, opened 2026-08-25
VCV-301: Add a zoom in zoom out reset button for images in the carousel

scho324:
SpecimenImageCarousel: 
* Added zooming in (+/-/reset) buttons next to the 'Image # of #' 
* Max zoom is 5x the default size, zooming in/out goes in increments of 0.5x 
* Added dragging to pan/move around the image while zoomed
* Added clamping to keep users from going too far off the edges of the image
* There should be cursor-grab/cursor-grabbing appearance when interacting with the zoomed image, but the cursor change isn't showing up for me; please test, and if it's not working we can get rid of it for now

SignedSpecimenImage: 
* Added style prop for image-dragging and zooming
* Added draggable=false to avoid the dragging doing anything to the image itself (otherwise it would've also tried to do some 'duplicate by dragging and dropping')

### DIFF
--- a/src/components/specimen/specimen-image-carousel.tsx
+++ b/src/components/specimen/specimen-image-carousel.tsx
@@ -257,7 +257,7 @@ export default function SpecimenImageCarousel({
                                                 : undefined
                                         }
                                         className={cn(
-                                            'bg-muted user-select:none relative flex aspect-4/3 items-center justify-center overflow-hidden rounded-lg border',
+                                            'bg-muted relative flex aspect-4/3 items-center justify-center overflow-hidden rounded-lg border select-none',
                                             zoomLevel > MIN_ZOOM &&
                                                 'touch-none',
                                         )}
--- a/src/components/specimen/specimen-image-carousel.tsx
+++ b/src/components/specimen/specimen-image-carousel.tsx
@@ -257,7 +257,7 @@ export default function SpecimenImageCarousel({
                                                 : undefined
                                         }
                                         className={cn(
-                                            'bg-muted relative flex aspect-4/3 items-center justify-center overflow-hidden rounded-lg border',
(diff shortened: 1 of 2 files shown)


src/components/specimen/specimen-image-carousel.tsx
```diff
@@ -148,19 +245,66 @@ export default function SpecimenImageCarousel({
...
-                                    />
-                                </div>
-                            </CarouselItem>
-                        ))}
+                        {allImagesForSpecimen.map((image, imageIndex) => {
+                            const isCurrentSlide =
+                                imageIndex === currentImageIndex;
+                            return (
+                                <CarouselItem key={image.id}>
+                                    <div
```
csmanel: I think this needs user-select:none, to prevent element selection highlighting

src/components/specimen/specimen-image-carousel.tsx
```diff
@@ -187,6 +331,38 @@ export default function SpecimenImageCarousel({
...
+                    >
+                        <RotateCcw className="h-4 w-4" />
+                        <span className="leading-none">Reset</span>
+                    </Button>
+                    <Button
+                        type="button"
+                        variant="outline"
+                        onClick={handleZoomIn}
+                        disabled={zoomLevel >= MAX_ZOOM}
+                        aria-label="Zoom out"
```
csmanel: I imagine this should be zoom in?

csmanel (commented):
two small things to address! otherwise looks good

src/components/specimen/specimen-image-carousel.tsx
csmanel: Sorry my mistake, this should be select-none not user-select:none.
Sorry!

csmanel (approved):

